PR #2, opened 2020-12-10
west: fix remote origin

sylvioalves:
Fixed python lint issues
Fixed requirements.txt due to build error


sylvioalves:
@nashif, it would make sense if at least me and @mahavirj could have approval access in here. How does that sound?

nashif:
> @nashif, it would make sense if at least me and @mahavirj could have approval access in here. How does that sound?

done

nashif (approved):

PR #3, opened 2020-12-14
west: add west installation

sylvioalves:
add esp-idf west python env environment

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -23,6 +23,9 @@ bitstring>=3.1.6
 ecdsa>=0.16.0
 python-socketio<5; python_version<="2.7"
 
+# add zephyr compability
+west>=0.8.0; python_version>="3.6"
+
 # windows-curses are required in Windows command line but cannot be installed in MSYS2. A requirement like
 # "windows-curses; sys_platform == 'win32'" would want to install the package on both of them. There is no environment
 # marker for detecting MSYS2. So instead, a dummy custom package is used with "windows-curses" dependency for Windows


sylvioalves:
@nashif, this fixes the issue you related.

nashif:
> @nashif, this fixes the issue you related.

not clear why we have to install those python packages in a separate script and why we need to do this here. I already have west 8.0 installed. Looks like this install them in the system rather in the user folder?

Should we maybe just rely on what is maintain in zephyr, i.e. in the requirements.txt of zephyr and add the things needed for esp to the zephyr requirement?

sylvioalves:
> > @nashif, this fixes the issue you related.
> 
> not clear why we have to install those python packages in a separate script and why we need to do this here. I already have west 8.0 installed. Looks like this install them in the system rather in the user folder?
> 
> Should we maybe just rely on what is maintain in zephyr, i.e. in the requirements.txt of zephyr and add the things needed for esp to the zephyr requirement?

@nashif, let me try to figure out another way to handle this if possible.

sylvioalves:
@nashif, I have been checking at slack channel whether it is possible to add esp-idf python dependencies in `scripts/requirements-xx.txt`. However, I think it might mess user environment and possibly break esp-idf builds.

The reason esp-idf uses virtual env is to have it's proper package versioning x compatibility. 
At this moment, I having west package as part of virtual env is the way to go in my opinion.

Do you have any other suggestion?

Thanks.

mahavirj:
@sylvioalves We can close this PR now, as we can work it around without using IDF py virtual env?

sylvioalves:
Removed IDF python packages dependency for ESP32 Zephyr support.
This #5 fixs it.

PR #10, opened 2021-02-03
esp_timer: add LAC timer support

sylvioalves:


mahavirj:
@sylvioalves 

- Should we revert earlier commit in `esp_timer` component (https://github.com/zephyrproject-rtos/hal_espressif/commits/zephyr/components/esp_timer) and then add support for LAC timer?
- Also it appears that additional commit to move wifi porting hal can be removed from this PR (part of other PR)?

sylvioalves:
@mahavir, PTAL

mahavirj (approved):

PR #14, opened 2021-02-25
esp32: counter: replaces CMakeLists.txt config macro name

glaubermaroto:
Zephyr community uses COUNTER instead of TIMER.
This commit replaces the config macro name to avoid confusion.

Signed-off-by: Glauber Maroto Ferreira <glauber.ferreira@espressif.com>


glaubermaroto:
PTAL @sylvioalves

sylvioalves (approved):

PR #26, opened 2021-05-26
wifi: event: move event handler to wifi driver

sylvioalves:
Remove esp_event API and handle wifi events
in wifi driver source


shubhamkulkarni97 (approved):
LGTM

PR #32, opened 2021-06-29
cmake: fix directory include

sylvioalves:
Twister is not finding subdir


uLipe (approved):
As a initial solution LGTM

glaubermaroto (approved):

PR #61, opened 2021-09-29
hal: wifi: add support of wifi subsystem for esp32c3

uLipe:
By adding wifi adapter and wpa_supplicant files.

Signed-off-by: Felipe Neves <felipe.neves@espressif.com>

### DIFF
--- a/components/esp_timer/src/esp_timer.c
+++ b/components/esp_timer/src/esp_timer.c
@@ -43,16 +43,12 @@ LOG_MODULE_REGISTER(LOG_MODULE_NAME);
 
 
 #if CONFIG_IDF_TARGET_ESP32
-#include "esp32/rom/ets_sys.h"
 #include "esp32/rtc.h"
 #elif CONFIG_IDF_TARGET_ESP32S2
-#include "esp32s2/rom/ets_sys.h"
 #include "esp32s2/rtc.h"
 #elif CONFIG_IDF_TARGET_ESP32S3
-#include "esp32s3/rom/ets_sys.h"
 #include "esp32s3/rtc.h"
 #elif CONFIG_IDF_TARGET_ESP32C3
-#include "esp32c3/rom/ets_sys.h"
 #include "esp32c3/rtc.h"
 #endif
 
--- a/components/esp_timer/src/esp_timer.c
+++ b/components/esp_timer/src/esp_timer.c
@@ -11,8 +11,6 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-
-#include <zephyr.h>
 #include <sys/param.h>
 #include <string.h>
 #include "soc/soc.h"
@@ -34,6 +32,8 @@
 #include "esp_timer_impl.h"
 #include "sdkconfig.h"
 
+#include <zephyr.h>
+
 #include "esp_private/startup_internal.h"
 #include "esp_private/esp_timer_private.h"
 #include "esp_private/system_internal.h"
@@ -43,12 +43,16 @@ LOG_MODULE_REGISTER(LOG_MODULE_NAME);
 
 
 #if CONFIG_IDF_TARGET_ESP32
+#include "esp32/rom/ets_sys.h"
 #include "esp32/rtc.h"
 #elif CONFIG_IDF_TARGET_ESP32S2
+#include "esp32s2/rom/ets_sys.h"
 #include "esp32s2/rtc.h"
 #elif CONFIG_IDF_TARGET_ESP32S3
+#include "esp32s3/rom/ets_sys.h"
 #include "esp32s3/rtc.h"
 #elif CONFIG_IDF_TARGET_ESP32C3
+#include "esp32c3/rom/ets_sys.h"
 #include "esp32c3/rtc.h"
 #endif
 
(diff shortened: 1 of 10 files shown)


uLipe:
@sylvioalves @glaubermaroto PTAL

zephyr/esp32c3/src/bt/esp_bt_adapter.c
```diff
@@ -31,7 +31,7 @@
 #include <zephyr.h>
 #include <sys/printk.h>
 #include <random/rand32.h>
-#include <drivers/interrupt_controller/intc_esp32c3.h>
+//#include <drivers/interrupt_controller/intc_esp32c3.h>
```
glaubermaroto, line 34: please, remove the commented line.

components/esp_timer/src/esp_timer.c
```diff
@@ -43,12 +43,16 @@ LOG_MODULE_REGISTER(LOG_MODULE_NAME);
...
+#include "esp32/rom/ets_sys.h"
 #include "esp32/rtc.h"
 #elif CONFIG_IDF_TARGET_ESP32S2
+#include "esp32s2/rom/ets_sys.h"
 #include "esp32s2/rtc.h"
 #elif CONFIG_IDF_TARGET_ESP32S3
+#include "esp32s3/rom/ets_sys.h"
 #include "esp32s3/rtc.h"
 #elif CONFIG_IDF_TARGET_ESP32C3
+#include "esp32c3/rom/ets_sys.h"
```
glaubermaroto, line 55: Is it necessary to include the same header to the other SoCs (aside esp32c3)? If so, why?

zephyr/esp32c3/src/common/esp_read_mac.c
```diff
@@ -0,0 +1,35 @@
...
+ * Copyright (c) 2021 Espressif Systems (Shanghai) Co., Ltd.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "esp_system.h"
+#include <rom/efuse.h>
+#include <zephyr.h>
+
+esp_err_t esp_read_mac(uint8_t *mac, esp_mac_type_t type)
```
glaubermaroto, line 11: For both esp32 and esp32s2, this functon is currently defined under `zephyr/esp_shared/src/common/esp_system_api.c`. Should we make the same for esp32c3?

uLipe: It is a bit different in the C3, I have plans of make this function common for it as well but in a later MR :)

components/esp_timer/src/esp_timer.c
uLipe: For C3 is necessary, for the others I just placed them just to keep the consistency.

zephyr/esp32c3/CMakeLists.txt
```diff
@@ -60,6 +67,10 @@ if(CONFIG_SOC_ESP32C3)
 
   zephyr_sources(
     ../../components/soc/esp32c3/gpio_periph.c
+    ../../components/esp_timer/src/ets_timer_legacy.c
+    ../../components/esp32c3/clk.c
```
sylvioalves, line 71: This file has been updated. Can you rebase with latest version and test if it still works?

components/esp_timer/src/esp_timer.c
glaubermaroto: If it is not necessary for the other SoCs, I suggest you consider not including it, otherwise, it may clutter the PR's history.

When another SoC needs this header, it can include in its own context. Does that make sense?

zephyr/esp32c3/src/common/esp_read_mac.c
glaubermaroto: Alternatively, you could consider unifying this function call in this same PR but in another commit, as done here:

https://github.com/zephyrproject-rtos/hal_espressif/pull/60/commits/9fcbf539faeee1220f57f56c82eda893f12912c6#diff-d3ff68eb35224ac623662cbaf1b15b2ca3d8131bc799c49b8c1550069ba79d3c

components/esp_timer/src/esp_timer.c
sylvioalves: Let me give my 5 cents in here: `esp32XX/rom/ets_sys.h` is not necessary at all. I think you can remove all entries in `esp_timer.c` regarding `ets_sys.h`. Can you re-check this?

zephyr/esp32c3/src/common/esp_read_mac.c
uLipe: I considered but it will take more time than just drop esp32c3 definitions, even the EFUSE addresses are not properly documented. I really prefer to do this in a separate PR later.

zephyr/esp32c3/src/bt/esp_bt_adapter.c
```diff
@@ -31,7 +31,6 @@
 #include <zephyr.h>
 #include <sys/printk.h>
 #include <random/rand32.h>
-#include <drivers/interrupt_controller/intc_esp32c3.h>
```
sylvioalves, line 34: OK, you removed it from here so that BT uses direct interrupt calls (as in WiFi), right?  
This could be a separate commit in this MR (touching BT stuff).

sylvioalves:
LGTM

zephyr/esp32c3/src/bt/esp_bt_adapter.c
uLipe: Yes, it will be added back when interrupt allocator gets merged.

sylvioalves (approved):

PR #80, opened 2021-11-17
hal: soc: esp32: cmake clean-up

glaubermaroto:
removes references to inexistent include directories.


sylvioalves (approved):
Have you tested against all samples?

glaubermaroto:
> Have you tested against all samples?

I've tested against hello_world, philosophers, WDT, counter (alarm), SPIRAM, and Wifi sample codes

PR #84, opened 2021-12-03
gpio: add rtc_io interface

sylvioalves:
This enables handling pins properly as some
gpios are configured over RTC registers.


sylvioalves:
@uLipe @glaubermaroto PTAL. ESP32-C3 does not need the patch.

uLipe (approved):

PR #85, opened 2022-01-06
flash: flash encryption support

sylvioalves:


components/spi_flash/esp32/flash_ops_esp32.c
```diff
@@ -81,7 +81,10 @@ esp_rom_spiflash_result_t IRAM_ATTR spi_flash_write_encrypted_chip(size_t dest_a
             break;
         }
     }
+#if defined(__ZEPHYR__)
+    memset(encrypt_buf, 0, sizeof(encrypt_buf));
+#else
     bzero(encrypt_buf, sizeof(encrypt_buf));
```
chrta, line 87: FYI (from "man bzero"):

> The bzero() function is deprecated (marked as LEGACY in POSIX.1-2001); use memset(3) in new programs.  POSIX.1-2008 removes the specification of bzero().  The bzero() function first appeared in 4.3BSD.

So it would be probably be also an option to remove bzero usage upstream (in esp-idf). See also https://github.com/espressif/esp-idf/issues/5095

uLipe (approved):

PR #86, opened 2022-01-08
uart: add signal sources map

sylvioalves:
This clears UART driver from keeping routing
signals hardcode in uart driver


glaubermaroto (approved):

sylvioalves:
@uLipe PTAL

uLipe (approved):

PR #93, opened 2022-02-22
pinctrl: script & pin states update

glaubermaroto:


dts/espressif/32s2/pinctrl.yaml
```diff
@@ -1,78 +1,111 @@
...
   sda:
     func: [in, out_high]
-    gpio: [[0, 21], [26, 45]]
     sigi: i2cext0_sda_in
     sigo: i2cext0_sda_out
-    outm: open_drain
+    gpio:
+      g0:
+        pins: [[0, 21], [26, 45]]
+        outm: open_drain
```
sylvioalves, line 85: Is is required to add pull-up option in I2C pins as well?

dts/espressif/32s2/esp32s2-pinctrl.dtsi
```diff
@@ -3147,7 +3147,6 @@
 
 			uart0_rx_gpio46: uart0_rx_gpio46 {
 				pinmux = <ESP32_PINMUX(46, ESP_U0RXD_IN, ESP_NOSIG)>;
-				bias-pull-up;
```
sylvioalves, line 3150: I know GPIO46 is input-only. Where is it written it has no PU support?

glaubermaroto: Here, it is not really about pull-up support on GPIO46 but rather that this pin is fixed to pull-down as per ESP32-S2's online documentation (check first note in [this section](https://docs.espressif.com/projects/esp-idf/en/latest/esp32s2/api-reference/peripherals/gpio.html#overview)).

sylvioalves (approved):

uLipe (approved):

PR #94, opened 2022-03-02
toolchain: changes to enable toolchain update

sylvioalves:


uLipe (approved):

components/log/include/esp_log.h
```diff
@@ -342,6 +346,7 @@ void esp_log_writev(esp_log_level_t level, const char* tag, const char* format,
         else                                { esp_log_write(ESP_LOG_INFO,       tag, LOG_SYSTEM_TIME_FORMAT(I, format), esp_log_system_timestamp(), tag, ##__VA_ARGS__); } \
     } while(0)
 #endif //CONFIG_LOG_TIMESTAMP_SOURCE_xxx
+*/
```
glaubermaroto, line 349: Why keep the original definition of ESP_LOG_LEVEL commented? 

Is ESP_LOG_LEVEL supposed to be defined based on some `ifdef` condition?

sylvioalves: The reason is that when it is enabled, the new toolchain causes `esp_log_write` function to fail due to wrong `%u` format. Also, we don't need ESP_LOG at all.

glaubermaroto: I see, consider removing it then.

glaubermaroto (approved):

PR #98, opened 2022-03-10
bt: adapter: fix missing nested lock check

sylvioalves:
Nested lock and unlock events handler was missing.

### DIFF
--- a/zephyr/esp32/src/bt/esp_bt_adapter.c
+++ b/zephyr/esp32/src/bt/esp_bt_adapter.c
@@ -338,6 +338,7 @@ static DRAM_ATTR int64_t s_time_phy_rf_just_enabled = 0;
 static DRAM_ATTR esp_bt_controller_status_t btdm_controller_status = ESP_BT_CONTROLLER_STATUS_IDLE;
 
 static unsigned int global_int_lock;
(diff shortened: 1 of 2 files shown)


uLipe (approved):

sylvioalves:
@glaubermaroto PTAL.

zephyr/esp32/src/bt/esp_bt_adapter.c
```diff
@@ -389,12 +390,20 @@ static void intr_on(unsigned int mask)
 
 static void IRAM_ATTR interrupt_disable(void)
 {	
-	global_int_lock = irq_lock();
+	if (global_nested_counter == 0) 
+		global_int_lock = irq_lock();
```
glaubermaroto, line 394: Not mandatory here, but should we keep this (and other occurrences) enclosed in curly braces to be consistent with Zephyr code style?

Remember we have agreed on using C89-style comments on hal_espressif the same way we do on Zephyr. What about extending that rule?

sylvioalves: yes!

glaubermaroto (approved):

PR #99, opened 2022-03-11
cmake: remove toolchain version check

sylvioalves:
This removes cmake gcc, ctng and crosstool check.
All those tools are not esp-idf related so this
check and warning messages can be removed.


components/esp32/project_include.cmake
```diff
@@ -26,6 +26,7 @@ endif()
 # Warn if the toolchain version doesn't match
 #
 # TODO: make these platform-specific for diff toolchains
-get_expected_ctng_version(expected_toolchain expected_gcc)
-gcc_version_check("${expected_gcc}")
-crosstool_version_check("${expected_toolchain}")
+# Zephyr: remove ctng, gcc and version check
+# get_expected_ctng_version(expected_toolchain expected_gcc)
+# gcc_version_check("${expected_gcc}")
+# crosstool_version_check("${expected_toolchain}")
```
glaubermaroto, line 32: What is the need to keep the original lines commented?

tools/cmake/idf.cmake
```diff
@@ -21,7 +21,8 @@ if(NOT __idf_env_set)
                             ${idf_path}")
         endif()
     else()
-        message(WARNING "IDF_PATH environment variable not found. Setting IDF_PATH to '${_idf_path}'.")
+        # Zephyr: remove IDF_PATH build warning
+        # message(WARNING "IDF_PATH environment variable not found. Setting IDF_PATH to '${_idf_path}'.")
```
glaubermaroto, line 25: Same question here.

components/esp32/project_include.cmake
sylvioalves: The only reason to keep original content commented is to make easy during next hal update. What do you think?

sylvioalves:
@uLipe PTAL.

components/esp32/project_include.cmake
glaubermaroto: I see, it is here to ease rebasing. My concern is that it potentially messes up the codebase in the mid/long run.

glaubermaroto (approved):

uLipe (approved):

PR #109, opened 2022-05-03
submodules: update repository to enable SDK integration

sylvioalves:


sylvioalves:
@uLipe PTAL in commits separately due to esptool and micro-ecc content.

uLipe (approved):

PR #111, opened 2022-05-05
blobs: removed all binary content from hal

sylvioalves:
As per from Zephyr policies, all .a and .bin files must be
removed from this repository.

All removed files are not part of Zephyr build and can be
excluded.


sylvioalves:
@stephanosio As pointed out, all binaries are now removed.

uLipe (approved):

stephanosio:
> @stephanosio As pointed out, all binaries are now removed.

@sylvioalves Thanks.

PR #125, opened 2022-08-23
blobs: update hal to handle blobs as from new policies

sylvioalves:


zephyr/module.yml
```diff
@@ -2,3 +2,260 @@ build:
   cmake: zephyr
   settings:
     dts_root: .
+blobs:
+  - path: lib/esp32c3/libbtbb.a
+    sha256: f77c12f9b0e607166523ecc44371a0cd4ec044183ae5c3292aa5b55eedbadbb9
+    type: lib
+    version: '1.0'
+    license-path: zephyr/blobs/license.txt
```
sylvioalves, line 10: @carlescufi, Can this licencing file be placed in here?

uLipe (approved):
Overall changes look good to me, I just would suggest to consider, on a future PR to proper document when to generate the blobs and how instead of placing it in the commit message for easy reference.

glaubermaroto (approved):

PR #127, opened 2022-08-25
compliance: drop all IRAM_ATTR from forward declarations

sylvioalves:
Make it compliant with Zephyr's, even that these files
are not build in CI.


sylvioalves:
@stephanosio, I guess this extends that PR by removing all forward declaration in this repository, at least at those files used in build.

PR #144, opened 2022-09-23
hal: pm: esp32_net: exclusion of clk source

glaubermaroto:
Removes source used for clock init when ESP32_NET is defined.

### DIFF
--- a/zephyr/esp32/CMakeLists.txt
+++ b/zephyr/esp32/CMakeLists.txt
@@ -162,6 +162,13 @@ if(CONFIG_SOC_ESP32 OR CONFIG_SOC_ESP32_NET)
     ../../components/hal/rtc_io_hal.c
     )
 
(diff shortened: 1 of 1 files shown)


zephyr/esp32/CMakeLists.txt
```diff
@@ -162,6 +162,13 @@ if(CONFIG_SOC_ESP32 OR CONFIG_SOC_ESP32_NET)
     ../../components/hal/rtc_io_hal.c
     )
 
+  if (NOT CONFIG_SOC_ESP32_NET)
+    zephyr_Sources(
```
LucasTambor, line 166: Why is there a capital S in sources?

glaubermaroto: thanks, have fixed that :)

sylvioalves (approved):

PR #149, opened 2022-10-26
driver: uart: allow esp32c3 usb-serial to work with wifi

sylvioalves:
ESP32-C3 usb-serial interface does not work when Wi-Fi is enabled. This fixes that by allowing hal_espressif to proper handle this pin when PHY is enabled.

### DIFF
--- a/components/esp_phy/src/phy_init.c
+++ b/components/esp_phy/src/phy_init.c
@@ -158,7 +158,7 @@ uint32_t IRAM_ATTR phy_enter_critical(void)
     }
 
     if(s_phy_lock_nest < 0xFFFFFFFF) {
-        s_phy_lock_nest++; 
+        s_phy_lock_nest++;
     }
 
     // Interrupt level will be stored in current tcb, so always return zero.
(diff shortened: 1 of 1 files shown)


components/esp_phy/src/phy_init.c
```diff
@@ -674,7 +674,7 @@ void esp_phy_load_cal_and_init(void)
     }
 #endif
 
-#if CONFIG_ESP_PHY_ENABLE_USB
+#if CONFIG_SERIAL_ESP32_USB
```
sylvioalves, line 677: phy init will only be called when BT or WiFi is enabled. In this case, if ESP32-C3USB port is used, this will automatically call the `phy_bbpll_en_usb`. This won't required another kconfig in Zephyr.

LucasTambor (approved):

components/esp_phy/src/phy_init.c
marekmatej: If the usb-serial is enabled elsewhere OR by default, this seems to be the correct fix.

marekmatej (approved):

PR #152, opened 2022-11-07
system: soc: Get console uart number from dts

marekmatej:
This will provide console number entirely from the DTS

Signed-off-by: Marek Matej <marek.matej@espressif.com>


marekmatej:
@sylvioalves, @LucasTambor PTAL

LucasTambor:
Neat solution for the UART_NUM problem! But there is one thing that makes me uncomfortable with this solution that comes back to Glauber's PR and I would like to bring back to discussion to hear from you. I don't think we should use the DT macros in hal_espressif. Maybe put them in soc.h or soc.c (we can use the CONFIG_ASSERT macro) under zephyr's repository maybe better suited for this kind of thing.

marekmatej:
> Neat solution for the UART_NUM problem! But there is one thing that makes me uncomfortable with this solution that comes back to Glauber's PR and I would like to bring back to discussion to hear from you. I don't think we should use the DT macros in hal_espressif. Maybe put them in soc.h or soc.c (we can use the CONFIG_ASSERT macro) under zephyr's repository maybe better suited for this kind of thing.

Attempt to implement the IDF configuration layer in the Zephyr `soc` is here:
https://github.com/zephyrproject-rtos/hal_espressif/pull/153
https://github.com/zephyrproject-rtos/zephyr/pull/52137

marekmatej:
PTAL folks

components/esp_system/port/soc/esp32/clk.c
```diff
@@ -28,7 +28,8 @@
 #else
 #include "stubs.h"
 #define CONFIG_ESP32_DEFAULT_CPU_FREQ_MHZ   ESP_SOC_DEFAULT_CPU_FREQ_MHZ
-#endif
+#define CONFIG_ESP32_RTC_CLK_CAL_CYCLES ESP_SOC_DEFAULT_RTC_CLK_CAL_CYCLES
```
sylvioalves, line 31: `#define CONFIG_ESP32_RTC_CLK_CAL_CYCLES ESP_SOC_DEFAULT_RTC_CLK_CAL_CYCLES` is already being defined in PM PR in Zephyr's upstream. Can you remove this entry from all `clk.c` files?

zephyr/esp_shared/include/stubs.h
```diff
@@ -1,26 +1,48 @@
 /*
- * Copyright (c) 2021 Espressif Systems (Shanghai) Co., Ltd.
+ * Copyright (c) 2022 Espressif Systems (Shanghai) Co., Ltd.
```
sylvioalves, line 2: You can keep `2021` in here as is.

sylvioalves (approved):

LucasTambor (approved):

PR #153, opened 2022-11-10
hal: esp32: move IDF configuration fragments to Zephyr

marekmatej:
Move IDF configuration fragments from hal_espressif to the Zephyr 'soc' where we can fetch configuration from the standart Zephyr sources, like devicetree.

Signed-off-by: Marek Matej <marek.matej@espressif.com>


zephyr/esp32c3/include/sdkconfig.h
```diff
@@ -4,13 +4,4 @@
...
-#define CONFIG_IDF_FIRMWARE_CHIP_ID 0x0005
-#define CONFIG_LOG_TIMESTAMP_SOURCE_RTOS 1
-#define CONFIG_BOOTLOADER_LOG_LEVEL 3
-
-#define CONFIG_BT_CTRL_ESP32C3 1
-#define CONFIG_BT_SOC_SUPPORT_5_0 1
-#define BTDM_CTRL_HCI_MODE_VHCI 1
-
-#define CONFIG_ESPTOOLPY_FLASHMODE_DIO 1
+#include "idf_config.h"
```
LucasTambor, line 7: Do we need the sdkconfig.h file at all?

marekmatej: Now is used to include the `idf_config.h` from `zephyr/soc`. We could find new place but the `sdkconfig.h` is spread around the IDF so it quite is convenient.

PR #158, opened 2022-11-25
pm: clock: check console status for proper flushing

sylvioalves:
Clock control needs to wait TX idle. However, it only needs to do that if console/uart is enabled.


marekmatej (approved):
LGTM

LucasTambor (approved):

PR #159, opened 2022-11-25
pm: remove uart baudrate check from its source

sylvioalves:
UART baudrate is already checked during drivers initialization.


marekmatej (approved):
seems legit

PR #160, opened 2022-11-27
system: soc: Add missing config

marekmatej:
When unifying stubs.h default config vanished.


sylvioalves:
Did we miss this or it is related to ADC/DAC support?

PR #166, opened 2022-12-21
stubs: fix default CPU clock definition

sylvioalves:
Retrieve CPU clock from DTS instead of custom CONFIG value. For ESP32 and ESP32-S2 where the CPU clock is linked to CCOUNT, it works as is. For ESP32-C3 where systimer is used, it does not work.

### DIFF
--- a/zephyr/esp_shared/include/stubs.h
+++ b/zephyr/esp_shared/include/stubs.h
@@ -16,7 +16,15 @@
 
 #include <zephyr/device.h>
 
-#define ESP_SOC_DEFAULT_CPU_FREQ_MHZ (CONFIG_SYS_CLOCK_HW_CYCLES_PER_SEC / 1000000)
+#if defined(CONFIG_SOC_ESP32) || defined(CONFIG_SOC_ESP32_NET)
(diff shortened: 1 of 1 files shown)


marekmatej (changes requested):
Better to replace `ESP_SOC_DEFAULT_CPU_FREQ_MHZ` by `CONFIG_SYS_CLOCK_HW_CYCLES_PER_SEC`

sylvioalves:
> CONFIG_SYS_CLOCK_HW_CYCLES_PER_SEC

I think it won't work. `CONFIG_SYS_CLOCK_HW_CYCLES_PER_SEC` is not always `cpu_clock` provided in device tree. In case o ESP32 and ESP32-S2, it is. For ESP32-C3, it is not. `CONFIG_SYS_CLOCK_HW_CYCLES_PER_SEC` in ESP32-C3 is the 16MHz timer clock reference. So better to keep this change.

sylvioalves:
@marekmatej @LucasTambor Any feedbacks?

LucasTambor (approved):

marekmatej (approved):
LGTM

PR #170, opened 2023-01-05
west: espressif monitor path

marekmatej:
Run the monitor using actual working directory, which is more likely to have build directory.  As we should be able to build anywhere in the west workspace.
Details in https://github.com/zephyrproject-rtos/zephyr/discussions/33521

### DIFF
--- a/west/tools.py
+++ b/west/tools.py
@@ -171,7 +171,7 @@ class Tools(WestCommand):
             esp_port = get_esp_serial_port(module_path)
 
         monitor_path = Path(module_path, "tools/idf_monitor.py")
-        cmd_path = Path(os.getenv("ZEPHYR_BASE")).absolute()
+        cmd_path = Path(os.getcwd())
         if platform.system() == 'Windows':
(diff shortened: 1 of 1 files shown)


west/tools.py
```diff
@@ -176,5 +176,6 @@ def monitor(self, module_path, args):
             cmd_exec(("python.exe", monitor_path, "-p", esp_port,
                      "-b", args.baud, elf_path, "--eol", args.eol), cwd=cmd_path)
         else:
+            cmd_path = Path(os.getcwd())
```
sylvioalves, line 179: 1) Does that work if you build an external project within zephyr folder?   
`<zephyrproject/zephyr>west build -b esp32 </home/xx/samples/heello_world> `

marekmatej: I've tested building in-tree and out-tree samples and it build & flash & run with monitor as expected. But I'd like anyone else to test it on their machines too.

sylvioalves (approved):

PR #180, opened 2023-02-02
Add support for esp32c3 hardware RS485 half duplex mode

dean-at-evos:
Changes required in the HAL to support half duplex mode for the esp32c3 only.

Signed-off-by: Dean Sellers <dsellers@evos.com.au>

**Draft**
This change only applies to the esp32C3, I am not familiar with the other SOC in this platform and haven't implemented or tested on these.


marekmatej:
As other ESP SOCs also support direct RS485, would you mind adding the pinctrl defs and ll funcs for them as well please?

dean-at-evos:
@marekmatej I can, although I won't be able to test it as I don't have any of the other SOC on hand. We are currently developing 3 products based on the ESP32C3, so that is the only hardware I have available to test on. If I can assume that the other variants work the same way, happy days! :)

dean-at-evos:
Adapting the changes here to all ESP SOC got messy. I have closed this PR in favour of creating a new one on a different branch that includes all SOC. See #185.

PR #186, opened 2023-02-20
components: esptool: Update to version 4.5

marekmatej:
Update esptool to fix issues with target reset
phase, which occurs on some host machines


sylvioalves:
@marekmatej, please, remove all those binaries (*.bin, etc..).

Is that possible and feasible to bring only the necessary/used files?

marekmatej:
> @marekmatej, please, remove all those binaries (*.bin, etc..).
> 
> Is that possible and feasible to bring only the necessary/used files?

Thanks, I did not spot the bin files. Removed all the `test` folder which contains the binaries and keys. Seems to work fine without it.

marekmatej:
@sylvioalves @LucasTambor  PTAL

PR #187, opened 2023-02-22
hal: esp32c3: soc_ll Software reset system

dean-at-evos:
Resetting via Zephyr sys_reboot() on ESP32C3 I was seeing intermittent hangs on restart. Changing the reset mode in the HAL from a processor level to system level reset solved this.

### DIFF
--- a/components/hal/esp32c3/include/hal/soc_ll.h
+++ b/components/hal/esp32c3/include/hal/soc_ll.h
@@ -44,7 +44,7 @@ static inline void soc_ll_unstall_core(int core)
 
 static inline void soc_ll_reset_core(int core)
 {
-    SET_PERI_REG_MASK(RTC_CNTL_OPTIONS0_REG, RTC_CNTL_SW_PROCPU_RST_M);
+    SET_PERI_REG_MASK(RTC_CNTL_OPTIONS0_REG, RTC_CNTL_SW_SYS_RST_M);
 }
 
 #ifdef __cplusplus


marekmatej (approved):
LGTM

marekmatej:
I have tested `soc_ll_reset_core()` with and with your changes and doesn't seem to make any difference. Both cpu and system reset flags makes it job. Can you provide better insight to the issue you are experiencing?

dean-at-evos:
I never really got to the bottom of why....but in our system which uses the on board ADC, UARTs and SPI2 in master mode sending the core only reset would sometimes (I just ran it again and got about 2 in 3 boots) result in the boot loader hanging at;
```
*** Booting Zephyr OS build zephyr-v3.3.0-60-g6a49efef0164 ***
[00:00:00.000,000] [0m<inf> main: Main started, 0, build time: Mar  6 2023 08:23:17[0m
ESP-ROM:esp32c3-api1-20210207
Build:Feb  7 2021
rst:0xc (RTC_SW_CPU_RST),boot:0xb (SPI_FAST_FLASH_BOOT)
Saved PC:0x40380d18
```
From this point only a power cycle reset will restart the system. I am running the MCUboot boot loader built and installed as per the instructions. Looking forward to the sysbuild in zephyr tree bootloader build :)
I suspect that a deeper dive would show that maybe any of the subsystems that we are using are changing something from the default which confuses the boot loader. The system reset however seems to solve the issue.
Since changing to the system reset I have had no hangs at reset at all. Just a quick test now, ran 10 resets from code and got a normal boot every time.
```
ESP-ROM:esp32c3-api1-20210207
Build:Feb  7 2021
rst:0x3 (RTC_SW_SYS_RST),boot:0xb (SPI_FAST_FLASH_BOOT)
Saved PC:0x40380d1c
SPIWP:0xee
mode:DIO, clock div:2
load:0x3fcd8598,len:0x12f0
load:0x403c8000,len:0x2c78
load:0x403d0000,len:0x163c
entry 0x403c804c
[esp32c3] [INF] [boot] chip revision: 3
[esp32c3] [INF] Enabling RNG early entropy source...
[esp32c3] [INF] *** Booting MCUboot build v1.9.0-162-g7453075 ***
[esp32c3] [INF] Primary image: magic=good, swap_type=0x2, copy_done=0x1, image_ok=0x1
[esp32c3] [INF] Scratch: magic=unset, swap_type=0x1, copy_done=0x3, image_ok=0x3
[esp32c3] [INF] Boot source: none
[esp32c3] [INF] Swap type: none
[esp32c3] [INF] Disabling RNG early entropy source...
[esp32c3] [INF] br_image_off = 0x10000
[esp32c3] [INF] ih_hdr_size = 0x20
[esp32c3] [INF] Loading image 0 - slot 0 from flash, area id: 1
[esp32c3] [INF] DRAM segment: start=0xae90, size=0xdf8, vaddr=0x3fc8c680
[esp32c3] [INF] IRAM segment: start=0x21a0, size=0x8cf0, vaddr=0x40380000
[esp32c3] [INF] start=0x40388984
*** Booting Zephyr OS build zephyr-v3.3.0-60-g6a49efef0164 ***
[00:00:00.000,000] [0m<inf> main: Main started, 0, build time: Mar  3 2023 17:12:07[0m
```
In my opinion calling a reset from code you would expect that CPU and subsystems are reset, but maybe there is a use case for just the core resetting....in which case a closer look at what is hanging the boot loader might be necessary.

marekmatej:
I see. For the ESP32-C3 case I think this is acceptable patch since it is only one core on the SoC.

dean-at-evos:
@marekmatej 
The TRM for the C3 chip (pg 177) defines the signal ```RTC_CNTL_SW_SYS_RST``` as a "Core Reset". The previously used ```RTC_CNTL_SW_PROCPU_RST``` is defined as "CPU Reset".
I know, I am confused too! I just really wanted a reliable system reset via software for remote sw updates. Maybe I'll use the WDT instead?
This function is called from https://github.com/zephyrproject-rtos/zephyr/blob/419dda3ed72d968e64d474994f13b7ecb623f18b/soc/riscv/esp32c3/soc.c which does some setup for the reset....
```
void IRAM_ATTR esp_restart_noos(void)
{
	/* Disable interrupts */
	csr_read_clear(mstatus, MSTATUS_MIE);

	/* Flush any data left in UART FIFOs */
	esp_rom_uart_tx_wait_idle(0);
	esp_rom_uart_tx_wait_idle(1);

	/* 2nd stage bootloader reconfigures SPI flash signals. */
	/* Reset them to the defaults expected by ROM */
	WRITE_PERI_REG(GPIO_FUNC0_IN_SEL_CFG_REG, 0x30);
	WRITE_PERI_REG(GPIO_FUNC1_IN_SEL_CFG_REG, 0x30);
	WRITE_PERI_REG(GPIO_FUNC2_IN_SEL_CFG_REG, 0x30);
	WRITE_PERI_REG(GPIO_FUNC3_IN_SEL_CFG_REG, 0x30);
	WRITE_PERI_REG(GPIO_FUNC4_IN_SEL_CFG_REG, 0x30);
	WRITE_PERI_REG(GPIO_FUNC5_IN_SEL_CFG_REG, 0x30);

	/* Reset wifi/bluetooth/ethernet/sdio (bb/mac) */
	SET_PERI_REG_MASK(SYSTEM_CORE_RST_EN_REG,
			SYSTEM_BB_RST | SYSTEM_FE_RST | SYSTEM_MAC_RST | SYSTEM_BT_RST |
			SYSTEM_BTMAC_RST | SYSTEM_SDIO_RST | SYSTEM_EMAC_RST |
			SYSTEM_MACPWR_RST | SYSTEM_RW_BTMAC_RST | SYSTEM_RW_BTLP_RST |
			BLE_REG_REST_BIT | BLE_PWR_REG_REST_BIT | BLE_BB_REG_REST_BIT);

	REG_WRITE(SYSTEM_CORE_RST_EN_REG, 0);

	/* Reset uart0 core first, then reset apb side. */
	SET_PERI_REG_MASK(UART_CLK_CONF_REG(0), UART_RST_CORE_M);

	/* Reset timer/spi/uart */
	SET_PERI_REG_MASK(SYSTEM_PERIP_RST_EN0_REG,
				SYSTEM_TIMERS_RST | SYSTEM_SPI01_RST | SYSTEM_UART_RST );
	REG_WRITE(SYSTEM_PERIP_RST_EN0_REG, 0);
	/* Reset dma */
	SET_PERI_REG_MASK(SYSTEM_PERIP_RST_EN1_REG, SYSTEM_DMA_RST);
	REG_WRITE(SYSTEM_PERIP_RST_EN1_REG, 0);

	/* Reset core */
	soc_ll_reset_core(0);

	while (true) {
		;
	}
}
```
This function resets some....not all peripherals??
During my debugging trying to find what was causing these hangs I didn't see them if I disabled my system writing to SPI. I figured it was the SPI2 not being reset here that caused the hang. Adding ```SYSTEM_SPI2_RST``` to the reset mask didn't seem to help though.

So?? Do you add in all the peripherals to the mask here to cover all use cases, or it is a combination of peripherials, and the order they are set/reset?

Or just call the system "Core reset", rather than "CPU Reset" which does that for you? I don't really mind which, it just seems more robust and easier do it at the SOC level.

sylvioalves (approved):

PR #193, opened 2023-03-06
esptool: add option to remove SHA256 digest

sylvioalves:
This fix brings from esp-idf a patch that
allows the bootloader binary not to append SHA256 digest when flash size is set to 'detect'.


marekmatej (approved):
Well done!

LucasTambor (approved):

PR #196, opened 2023-03-07
driver: uart_ll: fix wrong rs485 bit reference

sylvioalves:
RS485 support in ESP32-S3 uses different naming for some drivers struct entries. This fix it.


sylvioalves:
@dean-at-evos just a small fix from last RS485 support.

PR #194, opened 2023-03-06
bt: rework heap usage and improvements

sylvioalves:

PR #199, opened 2023-03-22
esp32c3: bt: adapter: fixup bt pwr on/off

twilfredo:
**Overview:**

This fixes an issue where when using bluetooth with the esp32c3 on Zephyr, the application would immediately hang after the boot banner.  This was introduced with https://github.com/zephyrproject-rtos/zephyr/commit/aba71499a9c59cce312d26cb0e54824bb922df1e. The problem is that in the call to [`esp_clk_init();`](https://github.com/zephyrproject-rtos/zephyr/blob/5a0834f16f1aaf8b204faa35272be43d8c8a0734/soc/riscv/esp32c3/soc.c#L88) ->  eventually ends up -> https://github.com/zephyrproject-rtos/hal_espressif/blob/8ee96508f329c7e59fcbf739f060cf12082dd805/components/esp_hw_support/port/esp32c3/rtc_init.c#L147
Where both wifi/bt power domains are turned off.  These need to be re-enabled to be able to use the hw.

**Testing**

This change was testing by running the `bluetooth/central` sample (which previously did not work) on the esp32c3-devkit.

**Notes:**

These changes are with reference to: https://github.com/zephyrproject-rtos/hal_espressif/blob/zephyr/components/bt/controller/esp32c3/bt.c.

Is this the correct approach?

**Fixes**
https://github.com/zephyrproject-rtos/zephyr/issues/55653
https://github.com/zephyrproject-rtos/zephyr/discussions/54131

### DIFF
--- a/zephyr/esp32c3/src/bt/esp_bt_adapter.c
+++ b/zephyr/esp32c3/src/bt/esp_bt_adapter.c
@@ -26,6 +26,7 @@
 #include "esp32c3/rom/ets_sys.h"
 #include <rom/efuse.h>
 #include <riscv/interrupt.h>
+#include "phy.h"
 
 #include <zephyr/kernel.h>
 #include <zephyr/sys/printk.h>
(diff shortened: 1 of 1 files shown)


sylvioalves:
@twilfredo, thnaks for bringing this. However. it has been already fixed in https://github.com/zephyrproject-rtos/hal_espressif/pull/194. Just waiting a final review to merge that. Do you mind if I close this PR?

twilfredo:
> @twilfredo, thnaks for bringing this. However. it has been already fixed in #194. Just waiting a final review to merge that. Do you mind if I close this PR?

Aha yep! that looks way better.

PR #202, opened 2023-03-22
cmake: add radio libraries check

sylvioalves:
This checks whether the radio binaries are downloaded when BT/BLE or Wi-Fi is enabled. In case these files are missing, cmake message and error and stops build. It also sets a explicit message describing download command.


sylvioalves:
@carlescufi, I've seen lots of question about build errors due to missing libraries. Even thought documentation says `west blobs fetch hal_espressif` is required, most people miss this. I've added this custom check for hal_espressif. Do you think there is anything we can do for all platforms somehow?

marekmatej (approved):

LucasTambor (approved):

PR #210, opened 2023-04-18
pm: esp32s3: Add support to power management and update related files

LucasTambor:
This commit introduces the power management support for esp32s3.
Add files to zephyr build and update IDF related files to version
4.4.4.

Update some RTC and power management related files to enable proper
configuration of sleep states. The previous implementation was preventing
future integration with esp32s3.

Adds a workaround calling esp_clk_private_lock() to take the lock before
stalling the other CPU.


zephyr/esp_shared/components/esp_pm/pm_impl.c
```diff
@@ -91,7 +91,6 @@
 #elif CONFIG_IDF_TARGET_ESP32S3
 /* Minimal divider at which REF_CLK_FREQ can be obtained */
 #define REF_CLK_DIV_MIN 2
-#define DEFAULT_CPU_FREQ CONFIG_ESP32S3_DEFAULT_CPU_FREQ_MHZ
```
sylvioalves, line 94: Isn't this needed at all?

PR #223, opened 2023-07-07
esp32s3: driver: add ADC support

sylvioalves:
Adds ADC source files for S3 SoC.

### DIFF
--- a/zephyr/esp32s3/src/esp_adc_cal/esp_adc_cal.c
+++ b/zephyr/esp32s3/src/esp_adc_cal/esp_adc_cal.c
@@ -1,0 +1,186 @@
+/*
+ * SPDX-FileCopyrightText: 2020-2021 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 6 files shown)


zephyr/esp32s3/src/esp_adc_cal/esp_adc_cal.c
```diff
@@ -0,0 +1,186 @@
...
+esp_adc_cal_value_t esp_adc_cal_characterize(adc_unit_t adc_num,
+        adc_atten_t atten,
+        adc_bits_width_t bit_width,
+        uint32_t default_vref,
+        esp_adc_cal_characteristics_t *chars)
+{
+    (void) default_vref;
+
+    // Check parameters
+    assert((adc_num == ADC_UNIT_1 || adc_num == ADC_UNIT_2, ESP_ADC_CAL_VAL_NOT_SUPPORTED));
```
EricNRS, line 135: I'm trying to merge your changes with PR 228.  Looks like you uses assert() here instead of returning error codes.  Is that something that we want to keep?

PR #224, opened 2023-07-14
driver: esp32: fix double promotions

XenuIsWatching:
Constants will get defined as doubles if missing the f at the end


components/driver/esp32c3/rtc_tempsensor.c
```diff
@@ -141,7 +141,7 @@ static float parse_temp_sensor_raw_value(uint32_t tsens_raw, const int dac_offse
     if (isnan(s_deltaT)) { //suggests that the value is not initialized
         read_delta_t_from_efuse();
     }
-    float result = (TSENS_ADC_FACTOR * (float)tsens_raw - TSENS_DAC_FACTOR * dac_offset - TSENS_SYS_OFFSET) - s_deltaT / 10.0;
```
keith-packard, line 144: Looks like the defined values `TSENS_ADC_FACTOR`, `TSENS_DAC_FACTOR`, and `TSENS_SYS_OFFSET` are all double constants that need 'f' appended too?

XenuIsWatching: you're right. done

keith-packard (approved):
These look fine. I note that there are more implicit double conversions in the i2c  and in other temp sensor driver bits (for esp32c3 at least). There are also lots in the test and example code, but I'm not sure Zephyr can even build that?

Are the computations in components/esp_rom/patches relevant? components/bt/host/bluedroid/external/sbc/plc?

How can we make sure we've got all of this fixed? Seems like piles of code here and I doubt much of it gets tested with the Zephyr test suite...

sylvioalves:
> These look fine. I note that there are more implicit double conversions in the i2c and in other temp sensor driver bits (for esp32c3 at least). There are also lots in the test and example code, but I'm not sure Zephyr can even build that?
> 
> Are the computations in components/esp_rom/patches relevant? components/bt/host/bluedroid/external/sbc/plc?
> 
> How can we make sure we've got all of this fixed? Seems like piles of code here and I doubt much of it gets tested with the Zephyr test suite...

@keith-packard, there are lots of unused code in there at this moment. We have been refactoring `hal_espressif` to bring latest esp-idf changes and also cleaning up unused stuff (by unused I mean in Zephyr level).

sylvioalves (approved):

PR #232, opened 2023-08-08
zephyr: esp32/c3/s2: compile power related files for poweroff

gmarull:
Some functionality guarded with CONFIG_PM is now needed by CONFIG_POWEROFF. While it is wrong to add such dependencies in a HAL, it allows us to move on for now. A fixme note has been added.

### DIFF
--- a/zephyr/esp32/CMakeLists.txt
+++ b/zephyr/esp32/CMakeLists.txt
@@ -161,15 +161,17 @@ if(CONFIG_SOC_SERIES_ESP32 OR CONFIG_SOC_SERIES_ESP32_NET)
     ../esp_shared/components/driver/rtc_io.c
     )
 
-  zephyr_sources_ifdef(
-    CONFIG_PM
-    ../esp_shared/components/driver/gpio.c
-    ../esp_shared/components/driver/rtc_io.c
-    ../esp_shared/components/esp_hw_support/sleep_modes.c
-    ../esp_shared/components/esp_pm/pm_impl.c
-    ../../components/esp_hw_support/port/esp32/rtc_sleep.c
-    ../../components/hal/rtc_io_hal.c
+  # FIXME: depending on Zephyr Kconfig options here is wrong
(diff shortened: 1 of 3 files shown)


sylvioalves (approved):

zephyr/esp32/CMakeLists.txt
```diff
@@ -161,15 +161,17 @@ if(CONFIG_SOC_SERIES_ESP32 OR CONFIG_SOC_SERIES_ESP32_NET)
...
 
-  zephyr_sources_ifdef(
-    CONFIG_PM
-    ../esp_shared/components/driver/gpio.c
-    ../esp_shared/components/driver/rtc_io.c
-    ../esp_shared/components/esp_hw_support/sleep_modes.c
-    ../esp_shared/components/esp_pm/pm_impl.c
-    ../../components/esp_hw_support/port/esp32/rtc_sleep.c
-    ../../components/hal/rtc_io_hal.c
+  # FIXME: depending on Zephyr Kconfig options here is wrong
```
sylvioalves, line 164: @gmarull, what is the recommended way? Move all dependencies to main repository?

gmarull: not really, dependencies must be explicit, e.g. CONFIG_ESP32_HAL_USE_RTC, when selected, it compiles rtc driver related code. Then, applications or drivers making use of such API explicitely select such symbol.

sylvioalves: So we need to create a few `ESP32_HAL_USE_xxxx` and re-work such file.Is that all required to meet the proper way?

gmarull: yes, IMHO can be done after this change if it's ok.

sylvioalves: Sure, I would not block this.

LucasTambor (approved):

PR #233, opened 2023-08-08
cmake: update source include to use ESP32_HAL_USE

sylvioalves:
This updates the way source files are added into build. Currently it's been using driver/peripherals Kconfig instead of a proper ESP32_USE_HAL entry.


zephyr/esp32/CMakeLists.txt
```diff
@@ -187,7 +187,7 @@ if(CONFIG_SOC_SERIES_ESP32 OR CONFIG_SOC_SERIES_ESP32_NET)
 
   endif()
 
-  if (CONFIG_MCUBOOT)
+  if (USE_ESP32_HAL_MCUBOOT)
```
gmarull, line 190: is this strictly for mcuboot?

zephyr/esp32s3/CMakeLists.txt
```diff
@@ -119,32 +119,32 @@ if(CONFIG_SOC_SERIES_ESP32S3)
...
     )
 
   zephyr_library_sources_ifdef(CONFIG_WDT_ESP32 ../../components/hal/wdt_hal_iram.c)
 
   zephyr_sources_ifdef(
-    CONFIG_PWM_LED_ESP32
+    CONFIG_USE_ESP32_HAL_LED_PWM
     ../../components/soc/esp32s3/ledc_periph.c
     ../../components/hal/ledc_hal_iram.c
     ../../components/hal/ledc_hal.c
```
gmarull, line 133: question: shouldn't soc/hal code be splitted (like e.g. STM32 LL/HAL layers)? I don't have knowledge to say here.

zephyr/esp32s3/CMakeLists.txt
```diff
@@ -119,32 +119,32 @@ if(CONFIG_SOC_SERIES_ESP32S3)
...
+    CONFIG_USE_ESP32_HAL_PCNT
     ../../components/hal/pcnt_hal.c
     )
 
   zephyr_sources_ifdef(
-    CONFIG_ADC_ESP32
+    CONFIG_USE_ESP32_HAL_ADC
     ../../components/hal/adc_hal.c
     ../esp_shared/components/driver/adc_common.c
     ../esp_shared/components/driver/rtc_io_hal.c
```
gmarull, line 150: ADC implies rtc stuff? looks wrong

zephyr/esp32/CMakeLists.txt
sylvioalves: Yes, it is. Should I keep `CONFIG_MCUBOOT`?

gmarull: it's ok, just was surprised HAL code specific to a particular bootloader. @nordicjm

zephyr/esp32s3/CMakeLists.txt
sylvioalves: `rtc_io` here is the subsystem responsible for handling low power and gpio configurations attach to RTC power domain. So it is indeed required.

zephyr/esp32s3/CMakeLists.txt
sylvioalves: Are you ok if we stick only with the new Kconfig entry? `hal_espressif` is really tied to esp-idf components and these kind of modification isn't accomplished in a snap.

zephyr/esp32s3/CMakeLists.txt
gmarull: so it should be ADC selecting it, no?

zephyr/esp32s3/CMakeLists.txt
gmarull: It depends if it later requires a re-write. if hal code using soc code? how does it work?

zephyr/esp32/CMakeLists.txt
nordicjm: @sylvioalves can you check mcuboot please? There is an issue with the espressif build on CI, RAM overflows by 5 bytes, so it is likely that this will be a problem for anyone wanting to use mcuboot

sylvioalves: @nordicjm, can you point the CI issue? I'll check right away.

nordicjm: See CI results on https://github.com/mcu-tools/mcuboot/pull/1774, thank you

PR #236, opened 2023-08-22
Add I2S pinctrl definition for ESP32C3 and ESP32S3

TiboDevC:
- Add I2S pinctrl definition for ESP32C3 and ESP32S3
- Compile _components/hal/i2s_hal.c_


components/hal/esp32c3/include/hal/clk_gate_ll.h
```diff
@@ -40,8 +40,8 @@ static inline uint32_t periph_ll_get_clk_en_mask(periph_module_t periph)
         return SYSTEM_UART1_CLK_EN;
     case PERIPH_I2C0_MODULE:
         return SYSTEM_I2C_EXT0_CLK_EN;
-    case PERIPH_I2S1_MODULE:
-        return SYSTEM_I2S1_CLK_EN;
+    case PERIPH_I2S0_MODULE:
```
sylvioalves, line 43: Although esp32c3 has a single I2S channel, esp-idf keeps its name referenced as "1".  
I suggest not changing it (and keep as is `I2S1`). PTAL: 
https://github.com/espressif/esp-idf/blob/master/components/hal/esp32c3/include/hal/clk_gate_ll.h#L38C10-L39C35  
https://github.com/espressif/esp-idf/blob/master/components/soc/esp32c3/include/soc/periph_defs.h

tools/cmake/idf.cmake
```diff
@@ -16,6 +16,8 @@ if(NOT __idf_env_set)
 
         if(NOT idf_path STREQUAL _idf_path)
             message(WARNING "IDF_PATH environment variable is different from inferred IDF_PATH.
+                            IDF_PATH environment variable: ${idf_path}
```
sylvioalves, line 19: This extra logging would not be needed in Zephyr as it automatically sets IDF_PATH. Did you have any issues prior this?

components/hal/esp32c3/include/hal/clk_gate_ll.h
TiboDevC: why esp-idf keeps 1? This is confusing, when reading it, we expect I2S0 as well

As esp-idf keeps 1, I agree to keep 1 in Zephyr

tools/cmake/idf.cmake
TiboDevC: I had an issue when setting up the environment and added this log to solve it.
I thought it could help others in the same situation. I had two idf folders and my environment was not coherent.

components/hal/esp32c3/include/hal/clk_gate_ll.h
TiboDevC: I pushed the change :heavy_check_mark:

tools/cmake/idf.cmake
TiboDevC: I removed this commit

sylvioalves:
BTW, would you squash all commits in a single one? I see no reason to not doing so.

TiboDevC:
Thanks for the review. The only reason is that I do not have an esp32s2 to test it, I can do it in another commit :)
Ok, I"ll squash them!

LucasTambor (approved):

marekmatej (approved):

PR #240, opened 2023-09-22
cmake: load all libc rom functions by default

sylvioalves:
RF libraries expects ROM functions as a form of
compatibility. Using different implementations can cause Wi-Fi/BLE errors/crashes. This PR adds a few libc functions to ESP32 environment.

### DIFF
--- a/components/esp_rom/esp32/ld/esp32.rom.newlib-funcs.ld
+++ b/components/esp_rom/esp32/ld/esp32.rom.newlib-funcs.ld
@@ -14,9 +14,9 @@
 
 abs = 0x40056340;
 __ascii_wctomb = 0x40058ef0;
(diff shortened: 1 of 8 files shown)


EricNRS:
Are there any issues reported that this fixes? Would like to know what the failures look like just in case that is part of the problems I am seeing on the ESP32-S3 WiFi.

sylvioalves:
> Are there any issues reported that this fixes? Would like to know what the failures look like just in case that is part of the problems I am seeing on the ESP32-S3 WiFi.

@EricNRS, it fixes #62473. However, I am still checking this PR among all test cases. You could give a try as well (but take only this commit, not the others).

LucasTambor (approved):

marekmatej (approved):

PR #243, opened 2023-09-23
esptool: Pad sections to account for alignment

keith-packard:
Parse ELF section alignment values and use those to check to see if successive sections are separated because of them. Pad a previous section if doing so would cause it to precisely align with the next section start address. This ensures that alignment requirements don't spoil our attempt to merge sections.


sylvioalves:
Hi @keith-packard, is this ready to be reviewed? By the way, what triggered you to submit this? Is that libc discussion or any other related issue?

sylvioalves (approved):

LucasTambor (approved):

PR #244, opened 2023-09-25
ESP32-S3 WiFi Add Private Heap

EricNRS:
driver: wifi: esp32s3: add private heap for WiFi

If packets are received faster than the system can process them, then the
system heap can be exhausted leading to system heap allocation errors and
system instability.

Add a private WiFi heap to limit allocation errors to the WiFi heap which
is handled gracefully within the WiFi library.

Fixes: 63043


zephyr/esp32s3/src/wifi/esp_wifi_adapter.c
```diff
@@ -36,6 +36,8 @@ LOG_MODULE_REGISTER(esp32_wifi_adapter, CONFIG_WIFI_LOG_LEVEL);
 #include "soc/dport_reg.h"
 #include "driver/adc2_wifi_private.h"
 
+#define WIFI_PKT_SIZE	1700
+K_HEAP_DEFINE(wifi_heap, WIFI_PKT_SIZE*MAX(CONFIG_NET_BUF_RX_COUNT, CONFIG_NET_BUF_TX_COUNT));
```
EricNRS, line 41: I added the WiFi packet size of 1700 based upon the allocation requests.  It seems that for a 1500 byte Ethernet packet, the allocation by the ESP WiFi library is 1700 bytes.

Ideally, we should allocate enough buffers for both RX and TX packets, but that is a large amount of memory and does not appear to improve performance in my testing.

zephyr/esp32s3/src/wifi/esp_wifi_adapter.c
```diff
@@ -36,7 +37,11 @@ LOG_MODULE_REGISTER(esp32_wifi_adapter, CONFIG_WIFI_LOG_LEVEL);
 #include "soc/dport_reg.h"
 #include "driver/adc2_wifi_private.h"
 
-K_THREAD_STACK_DEFINE(wifi_stack, 8192);
+#define WIFI_PKT_SIZE	1700
+K_HEAP_DEFINE(wifi_heap, WIFI_PKT_SIZE*MAX(CONFIG_NET_BUF_RX_COUNT, CONFIG_NET_BUF_TX_COUNT));
```
sylvioalves, line 41: The only issue I see is that it is duplicating RAM usage as this heap size is already allocated in net-stack (`net_pkt.c`). So adding this will increase RAM usage by a few 10~15kb (ok, depends on the configs).  Anyway, I understand this fixes the issue you have been facing. I will try to reproduce again to evaluate it.

EricNRS: Yes, the network packets and extra storage usage is not great, but it has always been the case -- it was just hidden in the system heap.  If you look at `esp32_wifi_send()`, it copies the TX packet into `data->frame_buf` to create a flat buffer and then that is passed to `esp_wifi_internal_tx()` which allocates a packet using `wifi_malloc()` and copies the data again.  Not ideal, but I didn't see any way to improve it. 

An alternative implementation to the `K_HEAP_DEFINE()` would be to allocate the `wifi_heap` from the system heap.  The issue I ran into is that if the SPIRAM heap is enabled, then this allocation is from SPIRAM and WiFi fails to connect.

zephyr/esp32s3/src/wifi/esp_wifi_adapter.c
EricNRS: This heap is also used to allocate the message queue and a few other OS items, so not sure if there should be a base allocation amount to cover that overhead or if those should be allocated from the system heap.  It may be that this buffer needs to be a separate Kconfig item with a range of buffers.

It could also be the case of thread priority issues that are causing the excessive buffer usage and maybe the best solution is to have a small buffer here and to fix the priority issues.  That would require some profiling and tracing which I don't have time for.  Let me change this PR to a draft while it gets discussed.

EricNRS:
@nashif @sylvioalves  Is the packet storm heap memory exhaustion handled differently now?  This fixes a legitimate system stability issue and potential DoS attack vector.

sylvioalves:
I should not have been closed.

vshymanskyy:
Please reopen

EricNRS:
No worries.  Once the HAL5.1 changes get merged, I will retest this and cleanup the patch and make it a KCONFIG option or similar.

I should have also noted that the heap exhaustion stability issue only applies if the available heap space is less that approximately 64KB (it is dependent upon the maximum buffer count and other allocations done in `esp_wifi_adapter.c`).

PR #249, opened 2023-11-23
system: make reset reason init public

sylvioalves:
Zephyr won't init constructor functions if CPP is disabled. By default, it is disabled.
This makes reset_reason_init public so Zephyr calls this during SoC startup.

### DIFF
--- a/components/esp_system/port/soc/esp32/reset_reason.c
+++ b/components/esp_system/port/soc/esp32/reset_reason.c
@@ -64,7 +64,7 @@ static esp_reset_reason_t get_reset_reason(uint32_t rtc_reset_reason, esp_reset_
     }
 }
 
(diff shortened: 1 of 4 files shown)


EricNRS:
Nice, that explains why it didn't work for me when I tested it a few months ago briefly, but then I did some more testing recently and it work fine.  Inbetween the two tests, I had enabled C++ support.

sylvioalves:
> Nice, that explains why it didn't work for me when I tested it a few months ago briefly, but then I did some more testing recently and it work fine. Inbetween the two tests, I had enabled C++ support.

Yes, I'd rather not enable CPP by default just because of it. So make it public. Will add the init call indo SoCs properly to initalize that.

LucasTambor (approved):

marekmatej (approved):

PR #255, opened 2023-12-06
components: Remove improper k_free() in esp_timer

peter-mitsis:
The thread object s_timer_task is statically allocated. Therefore its memory should not be freed via k_free().


sylvioalves:
@peter-mitsis, Hey,  thanks for submitting this. By the way, we have been working on new `hal_v5.1`, which will be available soon, and that wrong `k_free()` call was already fixed in there.  
https://github.com/zephyrproject-rtos/hal_espressif/blob/hal_v5.1/components/esp_timer/src/esp_timer.c

sylvioalves (approved):

marekmatej (approved):

PR #262, opened 2023-12-15
ble: fix task delete call

sylvioalves:
When BLE deinit is called, task delete is not
aborting and cleaning current internal task.


wmrsouza (approved):
LGTM

zephyr/esp32/src/wifi/esp_wifi_adapter.c
LucasTambor, line 1: `_queue_delete` is also using `delete_wrapper`. Would be better to create a `queue_delete_wrapper` that just frees the memory and rename the `delete wrapper` to `task_delete_wrapper`.

LucasTambor (changes requested):

marekmatej (approved):

wmrsouza (approved):

LucasTambor (approved):

PR #277, opened 2024-04-16
monitor: support using idf_monitor from a virtualenv on Windows

ajf58:
Ensure that the venv's Python executable is used, and convert the WindowsPath to a string.

### DIFF
--- a/west/tools.py
+++ b/west/tools.py
@@ -180,9 +180,5 @@ class Tools(WestCommand):
 
         monitor_path = Path(module_path, "tools/idf_monitor/idf_monitor.py")
         cmd_path = Path(os.getcwd())
-        if platform.system() == 'Windows':
-            cmd_exec((sys.executable, str(monitor_path), "-p", esp_port,
-                     "-b", args.baud, str(elf_path), "--eol", args.eol), cwd=cmd_path)
-        else:
-            cmd_exec((sys.executable, monitor_path, "-p", esp_port, "-b", args.baud,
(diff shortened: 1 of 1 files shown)


sylvioalves (approved):

west/tools.py
```diff
@@ -181,8 +181,8 @@ def monitor(self, module_path, args):
         monitor_path = Path(module_path, "tools/idf_monitor/idf_monitor.py")
         cmd_path = Path(os.getcwd())
         if platform.system() == 'Windows':
-            cmd_exec(("python.exe", monitor_path, "-p", esp_port,
-                     "-b", args.baud, elf_path, "--eol", args.eol), cwd=cmd_path)
+            cmd_exec((sys.executable, str(monitor_path), "-p", esp_port,
```
marekmatej, line 184: It almost looks like the `platform.system()` condition could be removed, right? Can you make the `cmd_exec` call universal for all platforms?

marekmatej (changes requested):
I'm wondering if we can unify both cases.

west/tools.py
ajf58: It might work, but I don't have a non-Windows host to try that change out with.

marekmatej: @ajf58 from what I see here, the only differences between Windows and non-Windows cases are the stringification of 2 arguments. Therefore I asked if you remove the condition. I'll test it here and let you know.

ajf58: OK, that's done. Please let me know how your testing gets on.

LucasTambor (approved):

PR #281, opened 2024-05-24
Restrict supported ESP32 chip revisions

marekmatej:
The ESP32 comes in a wide variety of chip and package revisions. Recent V3 revision brings important and major silicon and ROM fixes. To avoid isues, which comes from using the older revisions, we are restricting usage of older revisions, such as V0 and V1.

### DIFF
--- a/components/bootloader_support/src/bootloader_init.c
+++ b/components/bootloader_support/src/bootloader_init.c
@@ -45,6 +45,17 @@ esp_err_t bootloader_check_bootloader_validity(void)
     unsigned int major = revision / 100;
     unsigned int minor = revision % 100;
     ESP_EARLY_LOGI(TAG, "chip revision: v%d.%d", major, minor);
+
+#if defined(CONFIG_SOC_SERIES_ESP32)
+    if (major < 3) {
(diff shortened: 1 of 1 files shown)


wmrsouza (approved):

sylvioalves (approved):

components/bootloader_support/src/bootloader_init.c
```diff
@@ -45,6 +45,13 @@ esp_err_t bootloader_check_bootloader_validity(void)
     unsigned int major = revision / 100;
     unsigned int minor = revision % 100;
     ESP_EARLY_LOGI(TAG, "chip revision: v%d.%d", major, minor);
+
+#if defined(CONFIG_SOC_SERIES_ESP32) && !defined(CONFIG_ESP32_USE_UNSUPPORTED_REVISION)
+    if (major < 3) {
+        ESP_EARLY_LOGE(TAG, "Unsupported chip revision!");
```
sylvioalves, line 51: What about also adding something like:
```"This ESP32 SoC uses an unsupported revision. Please enable CONFIG_ESP32_USE_UNSUPPORTED_REVISION to allow its usage."```

sylvioalves: Even with the `CONFIG_ESP32_USE_UNSUPPORTED_REVISION=y`, I guess it can also print a warning message.

wmrsouza (approved):

PR #282, opened 2024-05-27
hal: esp32s3/esp32c3: add I2S support

wmrsouza:
added i2s support for esp32-s3 and esp32-c3

### DIFF
--- a/components/driver/deprecated/i2s_legacy.c
+++ b/components/driver/deprecated/i2s_legacy.c
@@ -327,7 +327,7 @@ static esp_err_t i2s_dma_intr_init(i2s_port_t i2s_num, int intr_flag)
 
     /* Set GDMA config */
     gdma_channel_alloc_config_t dma_cfg = {};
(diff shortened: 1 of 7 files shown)


LucasTambor (commented):
I think we should include the hal files and pincfg support only for S3 and C3. Since the support for other SoC won't be done right now

sylvioalves:
Do we need to change/touch `i2s_legacy.c` file? It is not used.

raffarost (approved):

wmrsouza:
> Do we need to change/touch `i2s_legacy.c` file? It is not used.

It really not used. I just mad
e changes to keep it compatible with modification made on `i2s_types.h` (to fix conflict with Zephyr types)

sylvioalves (approved):

sylvioalves:
@LucasTambor PTAL

marekmatej (approved):

raffarost (approved):

PR #290, opened 2024-06-25
esp-hal-components: esp_timer: fix recursive lock overwriting key

EricNRS:
A recursive lock was using the same key storage resulting the initial interrupt and system state being overwritten.

Removed the recursive lock and added an assert() to catch this in the future.

Fixes #74277

### DIFF
--- a/components/esp_timer/src/esp_timer.c
+++ b/components/esp_timer/src/esp_timer.c
@@ -165,12 +165,17 @@ esp_err_t IRAM_ATTR esp_timer_restart(esp_timer_handle_t timer, uint64_t timeout
         return ESP_ERR_INVALID_ARG;
     }
 
-    if (!is_initialized() || !timer_armed(timer)) {
+    if (!is_initialized()) {
         return ESP_ERR_INVALID_STATE;
     }
 
     esp_timer_dispatch_t dispatch_method = timer->flags & FL_ISR_DISPATCH_METHOD;
     timer_list_lock(dispatch_method);
+
+    if (!timer_armed(timer)) {
+        timer_list_unlock(dispatch_method);
+        return ESP_ERR_INVALID_STATE;
+    }
 
     const int64_t now = esp_timer_impl_get_time();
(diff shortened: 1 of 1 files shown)


EricNRS:
@sylvioalves - ready for review

components/esp_timer/src/esp_timer.c
```diff
@@ -165,13 +165,18 @@ esp_err_t IRAM_ATTR esp_timer_restart(esp_timer_handle_t timer, uint64_t timeout
...
-    if (!is_initialized() || !timer_armed(timer)) {
+    if (!is_initialized()) {
         return ESP_ERR_INVALID_STATE;
     }
 
     esp_timer_dispatch_t dispatch_method = timer->flags & FL_ISR_DISPATCH_METHOD;
     timer_list_lock(dispatch_method);
 
+    if (!timer_armed(timer)) {
+        timer_list_unlock(dispatch_method);
```
sylvioalves, line 176: @EricNRS, nice done! However, It seems we just need to remove the lock/unlock from `timer_remove` call. `timer_armed check could return to its original place as no locking would executed at line 173. I agree there is a double locking whenever `timer_remove` is called (2 places).

EricNRS: I was thinking there could be a race condition here where two threads could be trying to arm the same timer, so the `!timer_armed(timer)` check for both could succeed.  By moving the check inside the lock, only one will be successful.  I could see this happening if ISR context is one of the threads and the other thread is the timer thread or WiFi thread.

If you are sure this doesn't happen, I am happy to drop this change.

sylvioalves: Sounds good!

sylvioalves (approved):

marekmatej:
@EricNRS would you please squash those too commits please

marekmatej (approved):

sylvioalves:
@EricNRS I took the opportunity to squash the commits :)

PR #298, opened 2024-07-01
hal: esp32c6: Power management support

raffarost:
Add LL files for power management for C6

### DIFF
--- a/components/esp_hw_support/sleep_clock.c
+++ b/components/esp_hw_support/sleep_clock.c
@@ -15,10 +15,10 @@
 #include "esp_log.h"
 #include "esp_attr.h"
 #include "esp_check.h"
-#include "esp_regdma.h"
 #include "esp_private/startup_internal.h"
 #include "esp_private/sleep_retention.h"
 #include "esp_private/sleep_clock.h"
+#include "esp_regdma.h"
 
 #include "soc/pcr_reg.h"
 #include "modem/modem_syscon_reg.h"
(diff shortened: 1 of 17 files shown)


sylvioalves (approved):

sylvioalves:
@marekmatej @LucasTambor PTAL

components/esp_hw_support/sleep_modes.c
```diff
@@ -406,11 +403,11 @@ void esp_deep_sleep_deregister_hook(esp_deep_sleep_cb_t old_dslp_cb)
 // [refactor-todo] provide target logic for body of uart functions below
 static void IRAM_ATTR flush_uarts(void)
 {
-    for (int i = 0; i < SOC_UART_NUM; ++i) {
+    for (int i = 0; i < SOC_UART_HP_NUM; ++i) {
```
marekmatej, line 406: Does it mean that LP uarts are never suspended?

raffarost: I believe so. This is following current implementation for IDF

components/esp_hw_support/sleep_clock.c
LucasTambor, line 1: Is there a problem with the header order?

LucasTambor (approved):

components/esp_hw_support/sleep_clock.c
raffarost: This is to fix a duplicate definition of the symbol ARRAY_SIZE.
Changing the order fixes the error as `esp_regdma.h` does the check before defining the symbol again.

PR #306, opened 2024-07-23
soc: esp32c2: Add support to ESP32C2 and ESP8684

raffarost:
LL files and tools config update to support ESP32C2 and ESP8684


components/esp_hw_support/port/esp32c2/cpu_region_protect.c
```diff
@@ -86,7 +86,7 @@ void esp_cpu_configure_region_protection(void)
 
     if (esp_cpu_dbgr_is_attached()) {
         // Anti-FI check that cpu is really in ocd mode
-        ESP_FAULT_ASSERT(esp_cpu_dbgr_is_attached());
+//        ESP_FAULT_ASSERT(esp_cpu_dbgr_is_attached());
```
marekmatej, line 89: commented code should not be here

components/esp_hw_support/port/esp32c2/sar_periph_ctrl.c
```diff
@@ -24,7 +25,7 @@
 
 static const char *TAG = "sar_periph_ctrl";
 
-static int rtc_spinlock;
+int rtc_spinlock;
```
marekmatej, line 28: may I suggest to use `__` prefix in global symbols such as this lock?

components/hal/esp32c2/include/hal/clk_tree_ll.h
```diff
@@ -22,6 +22,7 @@
 extern "C" {
 #endif
 
+#undef MHZ
 #define MHZ                 (1000000)
```
marekmatej, line 26: I think, we have MHZ() macros in Zephyh somewhere, better to use them and not introduce new ones.

west/tools.py
```diff
@@ -31,6 +31,8 @@
 
 ESP_IDF_REMOTE = "https://github.com/zephyrproject-rtos/hal_espressif"
 
+build_elf_path = None
```
marekmatej, line 34: I highly suggest creating a separate PR for the changes in `west/tools.py` !

marekmatej (changes requested):

components/esp_hw_support/port/esp32c2/sar_periph_ctrl.c
raffarost: I think this will require changing code for other SoC's as well.
Should we do it afterwards in a separate PR?

components/hal/esp32c2/include/hal/clk_tree_ll.h
raffarost: Similar situation here. The conflict is between standard Zephyr macro and IDF macro.
Changing all IDF macros will affect all SoC's, requiring switching all IDF macros to Zephyr's.

components/esp_hw_support/port/esp32c2/sar_periph_ctrl.c
raffarost: We agreed this change should go in a separate PR, as it affects other SoC's as well.

components/hal/esp32c2/include/hal/clk_tree_ll.h
raffarost: We agreed this change should go in a separate PR, as it affects other SoC's as well.

PR #307, opened 2024-07-24
improvement: sdkconfig removal

wmrsouza:
Moves SoCs sdkconfig.h contents to a suitable Kconfig file


zephyr/Kconfig
```diff
@@ -1,3 +1,90 @@
-# Copyright 2022 Espressif Systems (Shanghai) PTE LTD
+# Copyright (c) 2024 Espressif Systems (Shanghai) Co., Ltd.
+# SPDX-License-Identifier: Apache-2.0
 
-# This file is required by Zephyr's build system and can be empty
+if MCUBOOT
```
sylvioalves, line 4: I think checking for `SOC_FAMILY_ESPRESSIF_ESP32` surrounding everything might make sense. Please check.

zephyr/Kconfig
```diff
@@ -1,3 +1,90 @@
...
+
+config EFUSE_MAX_BLK_LEN
+	int
+	default 192 if SOC_SERIES_ESP32
+	default 256 if SOC_SERIES_ESP32S2
+	default 256 if SOC_SERIES_ESP32S3
+	default 256 if SOC_SERIES_ESP32C3
+	default 256 if SOC_SERIES_ESP32C6
+
+#config SPI_FLASH_ROM_DRIVER_PATCH #remove? (it's also defined on zephyr/soc/espressif/commom/Kconfig.flash)
```
sylvioalves, line 30: I think it indeed could be hidden here as it is not something user can/should change.

sylvioalves: But it needs to be moved outside `MCUBOOT` sesction.

zephyr/Kconfig
```diff
@@ -1,3 +1,90 @@
...
+	bool
+	default y
+
+config ESP32_DPORT_DIS_INTERRUPT_LVL
+	int
+	default 5 if SOC_SERIES_ESP32 || SOC_SERIES_ESP32S3
+
+config FREERTOS_UNICORE
+	bool
+	default y if SOC_SERIES_ESP32
```
sylvioalves, line 50: It could be `default y if not SMP` only.

zephyr/esp32/include/sdkconfig.h
```diff
@@ -1,22 +1,7 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
  */
-#ifdef CONFIG_MCUBOOT
-#define CONFIG_EFUSE_VIRTUAL_OFFSET 0x250000
-#define CONFIG_EFUSE_VIRTUAL_SIZE 0x2000
-#define CONFIG_EFUSE_MAX_BLK_LEN 192
-#endif
 
 #define BOOTLOADER_BUILD 1
```
sylvioalves, line 7: Would be good to understand the issue when this is removed.

zephyr/Kconfig
wmrsouza: I'll remove it, since it is defined on zephyr/soc/espressif/commom/Kconfig.flash with default to y (independent of SOC_SERIES and bootloader option)

LucasTambor (approved):

zephyr/esp32/include/sdkconfig.h
wmrsouza: - when BOOTLOADER_BUILD is defined the implementation of bootloader_flash_read() calls bootloader_flash_read_allow_decrypt()
- when BOOTLOADER_BUILD is not defined the implementation of bootloader_flash_read() calls esp_flash_read()

marekmatej (approved):

PR #316, opened 2024-08-14
hal: add CAM HAL files

LucasTambor:
Add CAM files to provide a proper HAL API
for Zephyr's camera driver.

### DIFF
--- a/components/hal/cam_hal.c
+++ b/components/hal/cam_hal.c
@@ -1,0 +1,120 @@
+/*
+ * SPDX-FileCopyrightText: 2024 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "hal/cam_ll.h"
+#include "hal/cam_hal.h"
+#include "soc/soc_caps.h"
+
+/**
+ * @brief Configure line number to trigger interrupt
+ *
+ * @param hal CAM object data pointer
+ * @param num line number
+ *
+ * @return None
+ */
+static void cam_hal_set_line_int_num(cam_hal_context_t *hal, uint32_t num)
+{
+    if (num > 0) {
+        cam_ll_enable_hs_line_int(hal->hw, 1);
+        cam_ll_set_line_int_num(hal->hw, num);
+    } else {
+        cam_ll_enable_hs_line_int(hal->hw, 0);
+        cam_ll_set_line_int_num(hal->hw, 0);
+    }
+}
+
+/**
+ * @brief Configure V-SYNC filter threshold
+ *
+ * @param hal CAM object data pointer
+ * @param num V-SYNC filter threshold
+ *
+ * @return None
+ */
+static void cam_hal_set_vsync_filter_num(cam_hal_context_t *hal, uint32_t num)
+{
+    if (num > 0) {
+        cam_ll_enable_vsync_filter(hal->hw, 1);
+        cam_ll_set_vsync_filter_thres(hal->hw, num);
+    } else {
+        cam_ll_enable_vsync_filter(hal->hw, 0);
+        cam_ll_set_vsync_filter_thres(hal->hw, 0);
+    }
+}
+
+/**
+ * @brief Initialize CAM hardware
+ *
+ * @param hal    CAM object data pointer
+ * @param config CAM configuration
+ *
+ * @return None
+ */
+void cam_hal_init(cam_hal_context_t *hal, const cam_hal_config_t *config)
+{
+    memset(hal, 0, sizeof(cam_hal_context_t));
+
+    hal->hw = CAM_LL_GET_HW(0);
+
+    cam_ll_enable_stop_signal(hal->hw, 0);
+    cam_ll_swap_dma_data_byte_order(hal->hw, config->byte_swap_en);
+    cam_ll_reverse_dma_data_bit_order(hal->hw, 0);
+    cam_ll_enable_vsync_generate_eof(hal->hw, 1);
+
+    cam_hal_set_line_int_num(hal, 0);
+    cam_hal_set_vsync_filter_num(hal, 0);
+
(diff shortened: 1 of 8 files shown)


components/hal/esp32s3/include/hal/cam_ll.h
```diff
@@ -0,0 +1,638 @@
...
+
+/**
+ * @brief Start the CAM transaction
+ *
+ * @param dev CAM register base address
+ */
+__attribute__((always_inline))
+static inline void cam_ll_start(lcd_cam_dev_t *dev)
+{
+    dev->cam_ctrl.cam_update = 1;
```
epc-ake, line 518: Just a hint if something doesn't work out here...
I made the experience that the `cam_update` bit didn't clear itself, which lead to some strange behavior. 
Therefore I manually set it back to 0.

components/hal/cam_hal.c
```diff
@@ -0,0 +1,120 @@
...
+ * @param hal CAM object data pointer
+ *
+ * @return None
+ */
+void cam_hal_start_streaming(cam_hal_context_t *hal)
+{
+    cam_ll_reset(hal->hw);
+    cam_ll_fifo_reset(hal->hw);
+
+    cam_ll_start(hal->hw);
```
epc-ake, line 107: I intentionally started the dma after resetting the periphery but before starting the camera to make sure to not miss an incoming byte or capture an out going one. 
See [here](https://github.com/zephyrproject-rtos/zephyr/blob/8ebcb4534a7a5e161ed1b6d3e463ce19d3235e04/drivers/video/video_esp32_dvp.c#L191)

components/hal/esp32s3/include/hal/cam_ll.h
```diff
@@ -0,0 +1,638 @@
...
+ */
+static inline void cam_ll_enable_bus_clock(int group_id, bool en)
+{
+    (void)group_id;
+    SYSTEM.perip_clk_en1.lcd_cam_clk_en = en;
+}
+
+/// use a macro to wrap the function, force the caller to use it in a critical section
+/// the critical section needs to declare the __DECLARE_RCC_RC_ATOMIC_ENV variable in advance
+#define cam_ll_enable_bus_clock(...) (void)__DECLARE_RCC_RC_ATOMIC_ENV; cam_ll_enable_bus_clock(__VA_ARGS__)
```
epc-ake, line 38: I don't know what espressifs strategy is for the bus_clock controls but since these registers are not part of the cam_lcd pheriphery it might be worth considering to put them somewhere else?
In zephyr I used the `clock_control` subsystem. 
See [my implementation](https://github.com/zephyrproject-rtos/zephyr/blob/8ebcb4534a7a5e161ed1b6d3e463ce19d3235e04/drivers/clock_control/clock_control_esp32_cam.c#L34) and [clk_gate_ll.h](https://github.com/LucasTambor/hal_espressif/blob/ee9a5d0162046ddab5e63ed46511e46567bcfeb4/components/hal/esp32s3/include/hal/clk_gate_ll.h#L253)

epc-ake (commented):
nice work!
I think it shouldn't be a high effort to change the zephyr driver draft to use this API. 
The code snippets look very similar after all.

components/hal/cam_hal.c
```diff
@@ -0,0 +1,120 @@
...
+    cam_hal_set_line_int_num(hal, 0);
+    cam_hal_set_vsync_filter_num(hal, 0);
+
+    cam_ll_enable_invert_pclk(hal->hw, 0);
+    cam_ll_set_input_data_width(hal->hw, 8);
+    cam_ll_enable_invert_de(hal->hw, 0);
+    cam_ll_enable_invert_vsync(hal->hw, 0);
+    cam_ll_enable_invert_hsync(hal->hw, 0);
+    cam_ll_set_vh_de_mode(hal->hw, 0); // Disable vh_de mode default
+    cam_ll_enable_rgb_yuv_convert(hal->hw, 0); // bypass conv module default
```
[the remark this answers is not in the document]
LucasTambor: I totally understand what you are saying. The HAL file was taken from ESP-IDF v5.3. For reasons like maintenance and compliance is better if we try to stick with the original code. In Zephyr, nothing stops us from take this values from DTS and call the LL API to proper configure the registers.

components/hal/cam_hal.c
LucasTambor: What is the behaviour if we start dma before this routine?

components/hal/cam_hal.c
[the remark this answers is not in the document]
epc-ake: So you suggest that we ignore the `cam_hal_init` function and implement our own version in the Zephyr video driver? 
This would involve including `<hal/cam_ll.h>` and manually initializing the  `cam_hal_context_t` struct. 
I don't see an issue with this approach. I just want to ensure we're on the same page.

components/hal/cam_hal.c
epc-ake: I didn't try jet. I just fear that if the fifo's are not cleared before starting the dma that the dma starts to transfer some left over data.

components/hal/esp32s3/include/hal/cam_ll.h
```diff
@@ -0,0 +1,638 @@
...
+        break;
+    case LCD_CLK_SRC_PLL240M:
+        LCD_CAM.lcd_clock.lcd_clk_sel = 2;
+        break;
+    case LCD_CLK_SRC_XTAL:
+        LCD_CAM.lcd_clock.lcd_clk_sel = 1;
+        break;
+    default:
+        // disable LCD clock source
+        LCD_CAM.lcd_clock.lcd_clk_sel = 0;
```
epc-ake, line 107: Why use the lcd clock here and not `LCD_CAM.cam_ctrl.cam_clk*`

wmrsouza (approved):

sylvioalves (approved):

PR #329, opened 2024-08-27
zephyr: heap: Fix realloc call

marekmatej:
Use proper call and arguments to Zephyrs
system heap `z_realloc`.


LucasTambor (approved):

sylvioalves:
Folks, do we still need this `heap_caps` wrapper? I think this file could be removed instead.

sylvioalves (approved):

PR #331, opened 2024-09-02
Dynamic heap for ESP Wi-Fi

marekmatej:
Utilize the memory left unused by static allocation, and re-claim it dynamically using the Zephyr `k_heap` API for the ESP WiFi adapter purposes.

### DIFF
--- a/zephyr/esp32/CMakeLists.txt
+++ b/zephyr/esp32/CMakeLists.txt
@@ -4,7 +4,9 @@ if(CONFIG_SOC_SERIES_ESP32)
 
   zephyr_compile_options(-fstrict-volatile-bitfields)
   zephyr_compile_definitions_ifndef(CONFIG_MCUBOOT CONFIG_APP_BUILD_USE_FLASH_SECTIONS)
-  zephyr_compile_definitions_ifndef(CONFIG_MCUBOOT BOOTLOADER_BUILD=y) # See Zephyr-ESP / ZEP-820
+
+  zephyr_compile_definitions_ifdef(CONFIG_MCUBOOT BOOTLOADER_BUILD)
+  zephyr_compile_definitions_ifdef(CONFIG_ESP_SIMPLE_BOOT BOOTLOADER_BUILD)
 
   if(CONFIG_MCUBOOT)
     zephyr_compile_options(-fdump-rtl-expand)
@@ -242,12 +244,8 @@ if(CONFIG_SOC_SERIES_ESP32)
     )
   endif()
 
-  if (CONFIG_MCUBOOT)
-    zephyr_compile_definitions(BOOTLOADER_BUILD)
-    zephyr_compile_definitions(NDEBUG)
-  endif()
-
   if (CONFIG_MCUBOOT OR NOT CONFIG_BOOTLOADER_MCUBOOT)
+
     zephyr_include_directories(
       ../../components/esp_rom/${CONFIG_SOC_SERIES}
       ../port/include/boot
--- a/zephyr/esp32/src/wifi/esp_wifi_adapter.c
+++ b/zephyr/esp32/src/wifi/esp_wifi_adapter.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 Espressif Systems (Shanghai) Co., Ltd.
+ * Copyright (c) 2024 Espressif Systems (Shanghai) Co., Ltd.
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -10,9 +10,6 @@
 #include "zephyr_compat.h"
 
 #define CONFIG_POSIX_FS
-
-#include <zephyr/logging/log.h>
-LOG_MODULE_REGISTER(esp32_wifi_adapter, CONFIG_WIFI_LOG_LEVEL);
 
 #include "esp_wifi.h"
 #include "stdlib.h"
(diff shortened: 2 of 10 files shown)


zephyr/esp32/src/wifi/esp_wifi_adapter.c
```diff
@@ -37,6 +37,36 @@ LOG_MODULE_REGISTER(esp32_wifi_adapter, CONFIG_WIFI_LOG_LEVEL);
...
+
+static struct k_heap esp_wifi_heap;
+
+static int esp_wifi_heap_init(void)
+{
+	ESP_EARLY_LOGI(TAG, "ESP Wi-Fi heap init at 0x%x size %d kB.\n",
+			esp_wifi_heap_init_mem, ESP_WIFI_HEAP_SIZE/1024);
+	k_heap_init(&esp_wifi_heap, esp_wifi_heap_init_mem, ESP_WIFI_HEAP_SIZE);
+
+#if defined(CONFIG_WIFI) && defined(CONFIG_BT)
```
LucasTambor, line 59: Can't we do this in build time?

zephyr/esp32/src/wifi/esp_wifi_adapter.c
```diff
@@ -37,6 +37,36 @@ LOG_MODULE_REGISTER(esp32_wifi_adapter, CONFIG_WIFI_LOG_LEVEL);
...
+static void *esp_wifi_heap_init_mem = &z_mapped_end;
+#define ESP_WIFI_HEAP_SIZE  ((uintptr_t)&_heap_sentry - (uintptr_t)&z_mapped_end)
+
+static struct k_heap esp_wifi_heap;
+
+static int esp_wifi_heap_init(void)
+{
+	ESP_EARLY_LOGI(TAG, "ESP Wi-Fi heap init at 0x%x size %d kB.\n",
+			esp_wifi_heap_init_mem, ESP_WIFI_HEAP_SIZE/1024);
+	k_heap_init(&esp_wifi_heap, esp_wifi_heap_init_mem, ESP_WIFI_HEAP_SIZE);
```
LucasTambor, line 57: What if the extension does not have enough memory? Can we check and use sys_heap instead?

marekmatej: we don't know the `z_mapped_end` until the link phase is done.

marekmatej: > Can we check and use sys_heap instead?

Yes, we could add some kind of runtime mempool switching feature to the adapter to do this. Let's add this feature in another PR.

zephyr/esp32/src/wifi/esp_wifi_adapter.c
marekmatej: I don't think it is possible. At least it wouldn't be a straightforward solution.

LucasTambor: what about `BUILD_ASSERT()`?

marekmatej: > BUILD_ASSERT()

the expression behind the `ESP_WIFI_HEAP_SIZE` is not constant, same situation.

zephyr/esp32/src/wifi/esp_wifi_adapter.c
```diff
@@ -966,7 +1016,6 @@ esp_err_t esp_wifi_init(const wifi_init_config_t *config)
 	return result;
 }
 
-static void esp_wifi_free(void *mem)
-{
-	k_free(mem);
-}
+#ifdef CONFIG_ESP32_WIFI_HEAP_MEM_POOL
+SYS_INIT(esp_wifi_heap_init, PRE_KERNEL_2, 50);
```
sylvioalves, line 1020: I suggest replacing the hard-coded `50` level to `CONFIG_KERNEL_INIT_PRIORITY_DEVICE` instead.

zephyr/esp32s2/CMakeLists.txt
```diff
@@ -4,6 +4,9 @@ if(CONFIG_SOC_SERIES_ESP32S2)
 
   zephyr_compile_options(-fstrict-volatile-bitfields)
   zephyr_compile_definitions_ifndef(CONFIG_MCUBOOT CONFIG_APP_BUILD_USE_FLASH_SECTIONS)
+  zephyr_compile_definitions_ifdef(CONFIG_MCUBOOT BOOTLOADER_BUILD) # See Zephyr-ESP / ZEP-820
```
sylvioalves, line 7: What is the reason to include this (`NDEBUG`) only for ESP32-S2?
Would you also modifying all CMakeLists.txt to make these `BOOTLOADER_BUILD` definition in this same area? That would make it kind of common stuff among socs.

sylvioalves:
Overall LGTM. Is it possible for you to create a memory comparison when building `samples/net/wifi` with/without the new ESP_HEAP so we can evaluate the difference? Perhaps this info could be placed in `zephyr` upstream PR.

zephyr/esp32/src/wifi/esp_wifi_adapter.c
marekmatej: fixed

zephyr/esp32s2/CMakeLists.txt
marekmatej: yeah, this flag has been there since 4.1 hiding in the `sdkconfig.h`, and we keep copying it. I removed it and tested it with no issues now.

marekmatej:
> Overall LGTM. Is it possible for you to create a memory comparison when building `samples/net/wifi` with/without the new ESP_HEAP so we can evaluate the difference? Perhaps this info could be placed in `zephyr` upstream PR.

Zephyr PR description updated.

sylvioalves (approved):

zephyr/esp32/CMakeLists.txt
```diff
@@ -4,7 +4,9 @@ if(CONFIG_SOC_SERIES_ESP32)
 
   zephyr_compile_options(-fstrict-volatile-bitfields)
   zephyr_compile_definitions_ifndef(CONFIG_MCUBOOT CONFIG_APP_BUILD_USE_FLASH_SECTIONS)
-  zephyr_compile_definitions_ifndef(CONFIG_MCUBOOT BOOTLOADER_BUILD=y) # See Zephyr-ESP / ZEP-820
+
+  zephyr_compile_definitions_ifdef(CONFIG_MCUBOOT BOOTLOADER_BUILD) # See Zephyr-ESP / ZEP-820
```
LucasTambor, line 8: Should we put this kind of comment here?

marekmatej: fixed

wmrsouza (approved):

zephyr/esp32/CMakeLists.txt
```diff
@@ -4,7 +4,9 @@ if(CONFIG_SOC_SERIES_ESP32)
 
   zephyr_compile_options(-fstrict-volatile-bitfields)
   zephyr_compile_definitions_ifndef(CONFIG_MCUBOOT CONFIG_APP_BUILD_USE_FLASH_SECTIONS)
-  zephyr_compile_definitions_ifndef(CONFIG_MCUBOOT BOOTLOADER_BUILD=y) # See Zephyr-ESP / ZEP-820
+
+  zephyr_compile_definitions_ifdef(CONFIG_MCUBOOT BOOTLOADER_BUILD)
+  zephyr_compile_definitions_ifdef(CONFIG_ESP_SIMPLE_BOOT BOOTLOADER_BUILD)
```
sylvioalves, line 9: Having `CONFIG_ESP_SIMPLE_BOOT` here sounds wrong.

raffarost (approved):

LucasTambor (approved):

sylvioalves (approved):

zephyr/esp32/CMakeLists.txt
marekmatej: > Having `CONFIG_ESP_SIMPLE_BOOT` here sounds wrong. Why only for ESP32?

That is because the `BOOTLOADER_BUILD` is delimiting the DPORT access and that caused the startup issue, so we need to enable this flag here. Other SoCs don't have this kind of issue and no flag needs to be defined. However, we will remove that flag in the near future and this line will go away.

PR #333, opened 2024-09-07
flash: esp32: add cache handling during flash ops

sylvioalves:
Flash operations are not handling cache enable/disable operations, causing system crash due to interrupts placed in flash.


sylvioalves:
Converted to tag: https://github.com/zephyrproject-rtos/hal_espressif/releases/tag/bugfix_v3.7.1_flash_cache

PR #336, opened 2024-09-10
spi_flash: update initialization process and manage cache properly

sylvioalves:
Current implementation does not use spi-flash functions to disable/enable cache and interrupts during flash write operations. In order to enable those functions, spi-flash needs to be initialized during bootloader_init, supplying necessary chip information to it.

### DIFF
--- a/components/bootloader_support/src/esp32/bootloader_esp32.c
+++ b/components/bootloader_support/src/esp32/bootloader_esp32.c
@@ -211,6 +211,13 @@ esp_err_t bootloader_init(void)
     // print 2nd bootloader banner
     bootloader_print_banner();
 
+#ifndef CONFIG_BOOTLOADER_MCUBOOT
+    spi_flash_init_chip_state();
+    if ((ret = esp_flash_init_default_chip()) != ESP_OK) {
+        return ret;
+    }
+#endif
+
 #if !CONFIG_APP_BUILD_TYPE_RAM
     // reset MMU
     bootloader_reset_mmu();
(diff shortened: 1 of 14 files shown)


zephyr/esp32/CMakeLists.txt
```diff
@@ -147,6 +147,7 @@ if(CONFIG_SOC_SERIES_ESP32)
     ../../components/spi_flash/spi_flash_chip_th.c
     ../../components/spi_flash/spi_flash_chip_winbond.c
     ../../components/spi_flash/spi_flash_os_func_noos.c
```
LucasTambor, line 149: Is the `spi_flash_os_func_noos.c` beeing used now?

raffarost (approved):

components/bootloader_support/src/esp32/bootloader_esp32.c
```diff
@@ -211,6 +211,13 @@ esp_err_t bootloader_init(void)
     // print 2nd bootloader banner
     bootloader_print_banner();
 
+#ifdef CONFIG_ESP_SIMPLE_BOOT
```
marekmatej, line 214: Is this specific to simple-boot? Could it be used in MCUboot init as well?

zephyr/esp32/CMakeLists.txt
sylvioalves: Yes, it is used.

components/bootloader_support/src/esp32/bootloader_esp32.c
sylvioalves: It is used in mcuboot as well. 'CONFIG_ESP_SIMPLE_BOOT' is kept enabled for MCUBoot as well:
```c
config ESP_SIMPLE_BOOT
	bool "Simple Boot method"
	default y if !BOOTLOADER_MCUBOOT
	help
	  The Simple Boot is a method of booting that doesn't depend on a
	  2nd stage bootloader. Please note that some of the bootloader features
	  are not available using simple boot, such secure boot and OTA.
```

Perhaps we could check for `!BOOTLOADER_MCUBOOT` instead.

marekmatej: > Perhaps we could check for !BOOTLOADER_MCUBOOT instead.

Maybe `config ESP_SIMPLE_BOOT` should be fixed to exclude `MCUBOOT`. But in this case it is better to use `!BOOTLOADER_MCUBOOT`.

LucasTambor (approved):

wmrsouza (approved):

marekmatej (approved):

PR #361, opened 2024-11-06
esp32: fix naming of BT modem sleep orig config

real-tintin:
It seems like the naming of the config for BTDM modem sleep orig mode has changed name in esp-idf. In the latest doc (5.2.3) it is referred to as `CONFIG_BTDM_CTRL_MODEM_SLEEP_MODE_ORIG`, https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/kconfig.html#config-btdm-ctrl-modem-sleep-mode. I can find references to `CONFIG_BTDM_MODEM_SLEEP_MODE_ORIG` in 3.1, https://docs.espressif.com/projects/esp-idf/en/release-v3.1/api-reference/kconfig.html.

Furthermore, this has (now) an actual effect on modem mode. In a simple BLE sample where we advertise, the power consumption is reduced by over 50 % (from 70 to sub 30 mA on 3V).

### DIFF
--- a/zephyr/esp32/src/bt/esp_bt_adapter.c
+++ b/zephyr/esp32/src/bt/esp_bt_adapter.c
@@ -389,10 +389,10 @@ static DRAM_ATTR uint32_t btdm_lpcycle_us = 0;
 /* number of fractional bit for btdm_lpcycle_us */
 static DRAM_ATTR uint8_t btdm_lpcycle_us_frac = 0;
 
-#if CONFIG_BTDM_MODEM_SLEEP_MODE_ORIG
+#if CONFIG_BTDM_CTRL_MODEM_SLEEP_MODE_ORIG
 // used low power clock
 static DRAM_ATTR uint8_t btdm_lpclk_sel;
-#endif /* #ifdef CONFIG_BTDM_MODEM_SLEEP_MODE_ORIG */
+#endif /* #ifdef CONFIG_BTDM_CTRL_MODEM_SLEEP_MODE_ORIG */
 
 static DRAM_ATTR struct k_sem *s_wakeup_req_sem = NULL;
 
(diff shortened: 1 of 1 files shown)


sylvioalves:
@real-tintin, I don't se how this change has any effect at all. Also, `hal_espressif` is based on esp-idf v5.1.x, which does not have that change. We are planning to update do v5.3 soon.

real-tintin:
I hear you.

But let me share some actual results of the power consumption. These are some the simple BLE sample which I mentioned. BLE is configured and enabled, advertisement i started. All using the Zephyr BT API's.

**Without modem mode**
ble_sample_without_modem_sleep

**With modem mode**
ble_sample_with_modem_sleep

The device boots and renders a spike and about 2 s of higher power consumption. I guess this is the radio. Then the usage drops significantly, when entering modem sleep (avg 30 mA compared to 70 mA without sleep). The following spikes are the rate of advertisement, which here is fast mode (100-150 ms).

Correct me if I'm wrong, but *zephyr/esp32/src/bt/esp_bt_adapter.c* is not managed by esp-idf, it belongs to this project? In esp-idf there are no (longer) any mentions of `CONFIG_BTDM_MODEM_SLEEP_MODE_ORIG`, instead `CONFIG_BTDM_CTRL_MODEM_SLEEP_MODE_ORIG` is used, configurable *components/bt/controller/esp32/Kconfig.in*.

real-tintin:
I did some bisecting. The name change of the config in esp-idf occurred here:

```
commit 7c1c6697999b1f17eaeeb57bd09ae7db8a381cf5 (HEAD)
Author: zwj <zhiweijian@espressif.com>
Date:   Thu Jul 9 20:58:13 2020 +0800

    component/bt: add BLE v5.0 feature for bluedroid host
```

This is between their v4.2 and v4.3.

real-tintin:
> @sylvioalves 
> 
> I see the results you posted. How did you enable `CONFIG_BTDM_MODEM_SLEEP_MODE_ORIG` in there if there is no Zephyr's `kconfig` entry for `CONFIG_BTDM_MODEM_SLEEP_MODE_ORIG`? Did you manually enabled this in that file?
> 
> I think you are correct, we should have renamed `CONFIG_BTDM_MODEM_SLEEP_MODE_ORIG` to `CONFIG_BTDM_CTRL_MODEM_SLEEP_MODE_ORIG` (which we will).
> 
> I have a local PR that moves some of those BT/BLE `kconfig` entries to Zephyr's main repository. That will allow changing this configuration in the application itself.

I defined the configs in my application KConfig overlay, outside of Zephyr.

That sounds good. Better to define them in Zephyr. Let me know if I can help reviewing.

real-tintin:
@sylvioalves does there exist a related issue for this in Zephyr yet?

PR #369, opened 2024-11-14
esp32c6: add ble support

sylvioalves:
Brings necessary changes to enable BLE support.


guenterfischer:
Hi @sylvioalves
I tried to apply this patch to both 3.7.0 and 4.1.0, but it fails. What was the target version for that?
And is it planned to migrate this to the latest version of Zephyr?

PR #373, opened 2024-12-01
zephyr: port: add missing bootloader_wdt_feed

arifbalik:
mcuboot requires this function and it is present in the espressif port of the mcuboot bootloader. Also we already have the header file https://github.com/zephyrproject-rtos/hal_espressif/blob/zephyr/zephyr/port/include/boot/bootloader_wdt.h

https://github.com/mcu-tools/mcuboot/blob/main/boot/espressif/hal/src/bootloader_wdt.c

### DIFF
--- a/zephyr/esp32/CMakeLists.txt
+++ b/zephyr/esp32/CMakeLists.txt
@@ -276,6 +276,7 @@ if(CONFIG_SOC_SERIES_ESP32)
     zephyr_sources_ifdef(
       CONFIG_MCUBOOT
       ../port/boot/esp_image_loader.c
+      ../port/boot/bootloader_wdt.c
       )
   endif()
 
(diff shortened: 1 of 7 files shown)


sylvioalves:
@arifbalik Are you facing build or runtime issues? Will check this tomorrow. Thanks for fixing!

arifbalik:
> @arifbalik Are you facing build or runtime issues? Will check this tomorrow. Thanks for fixing!

Yes, spesifically this line fails because zephyr port has the header but no implementation;

https://github.com/mcu-tools/mcuboot/blob/742978e7b5c0c51c0cdeed6d5dba524978fdcd02/boot/espressif/hal/include/mcuboot_config/mcuboot_config.h#L188

sylvioalves:
I could not reproduce any issue when either building MCUBoot Zephyr's por nor Espressif Port.
When building the Espressif Port, are you setting `-DESP_HAL_PATH=` to a esp-idf v5.x path or to this `hal_espressif` repoditory?

arifbalik:
Ah my bad, turns out this was fixed in 4.0.0, I was working with 3.7.0

PR #379, opened 2024-12-11
Add IEEE 802.15.4 support

martinjaeger:
Draft PR to show current status of porting to Zephyr.

Code compiles, but is not yet tested.

@sylvioalves There are some files which are not needed for Zephyr. How do you typically handle this? Should the entire component folder be copied so that it's easier to compare it to upstream ESP-IDF files or should we just use files required by Zephyr?


marekmatej:
> How do you typically handle this? Should the entire component folder be copied so that it's easier to compare it to upstream ESP-IDF files or should we just use files required by Zephyr?

This is usually done by this repo https://github.com/espressif/esp-hal-3rdparty by fetching only chosen components. The rest of the work (like you did here) is done in the `hal_espressif`.

marekmatej:
@martinjaeger We are using these adaptation layers for BT and WiFi (`zephyr/esp32c6/src/wifi/esp_wifi_adapter.c`, `zephyr/esp32c3/src/bt/esp_bt_adapter.c`). Have you considered creating one for 802154?

martinjaeger:
> @martinjaeger We are using these adaptation layers for BT and WiFi (`zephyr/esp32c6/src/wifi/esp_wifi_adapter.c`, `zephyr/esp32c3/src/bt/esp_bt_adapter.c`). Have you considered creating one for 802154?

I prefer to avoid chip-specific adapters if possible. The changes required to compile the original ESP-IDF component for Zephyr look fairly small, so I hope we can keep the IEEE 802.15.4 implementation generic (same for all Espressif chips).

Rebased and fixed missing timer initialization.

PR #383, opened 2024-12-20
wifi: isr: use interrupt allocator

sylvioalves:
Make sure Wi-Fi adapter uses the interrupt allocator so that it won't mess with other peripheral's interrupt.

### DIFF
--- a/zephyr/esp32/src/wifi/esp_wifi_adapter.c
+++ b/zephyr/esp32/src/wifi/esp_wifi_adapter.c
@@ -32,6 +32,7 @@
 #include "wifi/wifi_event.h"
 #include "esp_private/adc_share_hw_ctrl.h"
 #include "esp_heap_runtime.h"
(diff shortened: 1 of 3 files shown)


zephyr/esp32/src/wifi/esp_wifi_adapter.c
```diff
@@ -523,8 +524,7 @@ static void clear_intr_wrapper(uint32_t intr_source, uint32_t intr_num)
 
 static void set_isr_wrapper(int32_t n, void *f, void *arg)
 {
-	irq_disable(0);
-	irq_connect_dynamic(0, n, f, arg, 0);
+	esp_intr_alloc(n, 0, f, arg, NULL);
```
sylvioalves, line 527: No need to call for source 2 as done in other SoCs.

wmrsouza (approved):

raffarost (approved):

LucasTambor (approved):

PR #392, opened 2025-01-08
adc: fix for crashes due to multiple calls to adc#_config_channel_atten()

wmrsouza:
It was observed that multiple calls to the adc#_config_channel_atten() functions caused execution crashes.
To avoid such crashes , this PR removes adc_hal_calibration_init() calls from adc#_config_channel_atten() functions, since adc_hal_calibration_init() is already called during Zephyr adc driver instantiation.


LucasTambor (approved):

sylvioalves (approved):
It is worth mentioning that the calibration is already being executed during adc initialization.

marekmatej (approved):

PR #399, opened 2025-01-22
zephyr: scripts: fix partition generator

marekmatej:
Update the partition scheme by adding variable partition sizes.
Added labels to all partition schemes as they are vital to the code.

### DIFF
--- a/zephyr/scripts/partitions/esp_genpartition.py
+++ b/zephyr/scripts/partitions/esp_genpartition.py
@@ -2,158 +2,121 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import sys
+import os 
 from datetime import datetime
-from collections import namedtuple
 
-# where the files ought to be copied
-output_path_prefix = "dts/common/espressif/"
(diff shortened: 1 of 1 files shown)


raffarost (commented):
just a thought: can hardcoded values be calculated or commonized somehow as defines in the file's header, for easy maintenance? also wondering about the source of the values, whether they could be derived from a main common source

marekmatej:
> just a thought: can hardcoded values be calculated or commonized somehow as defines in the file's header, for easy maintenance? also wondering about the source of the values, whether they could be derived from a main common source (for safety)

That was my idea too. However, that would increase the complexity and could bring issues. I decide to go with this simplistic version. To design a partition layout/sizes I use the spreadsheet, then feed the script and it works fine. We can for sure update it when we find robust solution.

raffarost (approved):

wmrsouza (approved):

sylvioalves (approved):

raffarost (approved):

PR #420, opened 2025-02-19
esp_rom_flash: handle read/write alignment

sylvioalves:
Check alignment before proceeding with operation.
If src, dest or length are not 4-byte align, use a temp buffer to handle it.

### DIFF
--- a/zephyr/port/bootloader/bootloader_flash.c
+++ b/zephyr/port/bootloader/bootloader_flash.c
@@ -30,6 +30,13 @@
 
 #include "esp_rom_spiflash.h"
 
+#include <zephyr/kernel.h>
+#include <zephyr/device.h>
+
+#define DT_DRV_COMPAT espressif_esp32_flash_controller
+#define SOC_NV_FLASH_NODE DT_INST(0, soc_nv_flash)
+#define FLASH_BLK_SZ DT_PROP(SOC_NV_FLASH_NODE, write_block_size)
+
 #ifdef CONFIG_EFUSE_VIRTUAL_KEEP_IN_FLASH
 #define ENCRYPTION_IS_VIRTUAL (!efuse_hal_flash_encryption_enabled())
 #else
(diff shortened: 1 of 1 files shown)


zephyr/port/bootloader/bootloader_flash.c
```diff
@@ -30,6 +30,13 @@
 
 #include "esp_rom_spiflash.h"
 
+#include <zephyr/kernel.h>
+#include <zephyr/device.h>
+
+#define DT_DRV_COMPAT espressif_esp32_flash_controller
+#define SOC_NV_FLASH_NODE DT_INST(0, soc_nv_flash)
+#define FLASH_BLK_SZ DT_PROP(SOC_NV_FLASH_NODE, write_block_size)
```
sylvioalves, line 38: This is usually small buffer.. up to 32 bytes. We could increase it to 256 byte-long as well.

sylvioalves:
OTA tests ok!

wmrsouza (approved):

zephyr/port/bootloader/bootloader_flash.c
```diff
@@ -333,53 +340,93 @@ static esp_err_t bootloader_flash_read_allow_decrypt(size_t src_addr, void *dest
...
-    if (allow_decrypt) {
-        return bootloader_flash_read_allow_decrypt(src_addr, dest, size);
-    } else {
-        return bootloader_flash_read_no_decrypt(src_addr, dest, size);
+    /* handle misaligned source address */
+    size_t src_offset = src_addr & 3;
+    size_t start_addr = src_addr - src_offset;
+
+    while (remaining > 0) {
+        copy_size = (remaining > FLASH_BLK_SZ) ? FLASH_BLK_SZ : remaining;
```
martinjaeger, line 353: This should use `sizeof`, just in case the `temp_buf`'s size is changed in the future.

```suggestion
        copy_size = (remaining > sizeof(temp_buf)) ? sizeof(temp_buf) : remaining;
```

zephyr/port/bootloader/bootloader_flash.c
```diff
@@ -333,53 +340,93 @@ static esp_err_t bootloader_flash_read_allow_decrypt(size_t src_addr, void *dest
...
+        aligned_size = (copy_size + src_offset + 3) & ~3;
+
+        /* read from flash into the temp buffer */
+        esp_err_t err;
+        if (allow_decrypt) {
+            err = bootloader_flash_read_allow_decrypt(start_addr, temp_buf, aligned_size);
```
martinjaeger, line 359: `aligned_size` could be more than the buffer size, leading to buffer overflow.

Example (in first run of the loop):
```
src_addr = 3;
src_offset = 3;
start_addr = 0;
copy_size = 32;
aligned_size = (32 + 3 + 3) & ~3 = 36;
```

martinjaeger (changes requested):
I think we need some tests to check these alignment calculations. It's quite difficult to be absolutely sure that all edge cases are covered just by looking at the code.

If I didn't get it wrong, the current implementation might again lead to a buffer overflow.

zephyr/port/bootloader/bootloader_flash.c
sylvioalves: Perhaps we can add the max offset which would be 3.. then rouding to 4:
`uint8_t temp_buf[FLASH_BLK_SZ + 4];`

martinjaeger:
I'm wondering why we actually need an implementation that allows unaligned writes. I've looked at the STM32 driver in Zephyr upstream and it would just return an error code if alignment is not correct. Can't we do the same here and avoid the difficult automatic alignment implementation?

PR #335, opened 2024-09-10
sync: update sources to meet hal updates

sylvioalves:

PR #438, opened 2025-05-08
hal: esp32c6: Fix sleep routine

raffarost:
Fix sleep routine for ESP32-C6 after regression when updating HAL to v.5.1.c (PR #335).

### DIFF
--- a/components/esp_hw_support/sleep_modes.c
+++ b/components/esp_hw_support/sleep_modes.c
@@ -522,11 +522,11 @@ static void IRAM_ATTR resume_timers(uint32_t pd_flags) {
 // [refactor-todo] provide target logic for body of uart functions below
 static void IRAM_ATTR flush_uarts(void)
 {
-    for (int i = 0; i < SOC_UART_NUM; ++i) {
+    for (int i = 0; i < SOC_UART_HP_NUM; ++i) {
 #ifdef CONFIG_IDF_TARGET_ESP32
         esp_rom_uart_tx_wait_idle(i);
 #else
-        if (periph_ll_periph_enabled(PERIPH_UART0_MODULE + i)) {
+        if (uart_ll_is_enabled(i)) {
             esp_rom_uart_tx_wait_idle(i);
         }
 #endif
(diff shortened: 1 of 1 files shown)


components/hal/esp32c6/include/hal/uart_ll.h
```diff
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2022-2024 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2022-2023 Espressif Systems (Shanghai) CO LTD
```
raffarost, line 2: Date wasn't update as file is a copy from previous HAL version.

components/hal/esp32c6/include/hal/uart_ll.h
wmrsouza, line 1: Why not all `FORCE_INLINE_ATTR` were replaced by `static inline`?

wmrsouza: Why some `HAL_FORCE_` macros were replaced by expanded version and others not?

raffarost: This file was copied as is from IDF, and is being reverted back before v.5.1.4.
I'll address your comments to improve the file.

LucasTambor, line 1: You are reverting some changes that were not added by https://github.com/zephyrproject-rtos/hal_espressif/pull/335 but that are needed for esp32c6's LP Core support. Did you tested this with LP Core build?

LucasTambor (changes requested):

components/hal/esp32c6/include/hal/uart_ll.h
raffarost: Thanks for pointing it out. I checked IDF's individual changes and probably doing the minimal changes is a better option.

sylvioalves:
@marekmatej

sylvioalves (approved):

sylvioalves (approved):

PR #445, opened 2025-05-24
espressif: mcuboot: get MCUBOOT_BOOT_MAX_ALIGN value from DT

almir-okato:
- Get MCUBOOT_BOOT_MAX_ALIGN value from DT - This change affects only MCUboot Zephyr Port for ESP chips or Zephyr application with MCUboot compatibilty (which builds the bootutil lib).

- Wrap typo fix with ifdef for MCUboot Espressif Port compatibility - The typo fix is not present on the original version of ESP-IDF used by MCUboot, thus the partial reverting of the fix.

### DIFF
--- a/zephyr/port/include/boot/mcuboot_config/mcuboot_config.h
+++ b/zephyr/port/include/boot/mcuboot_config/mcuboot_config.h
@@ -7,6 +7,8 @@
 #ifndef __MCUBOOT_CONFIG_H__
 #define __MCUBOOT_CONFIG_H__
 
+#include <zephyr/devicetree.h>
+
 /*
  * Signature types
  *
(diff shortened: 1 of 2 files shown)


zephyr/port/include/boot/mcuboot_config/mcuboot_config.h
```diff
@@ -7,6 +7,8 @@
 #ifndef __MCUBOOT_CONFIG_H__
 #define __MCUBOOT_CONFIG_H__
 
+#include <zephyr/devicetree.h>
```
marekmatej, line 10: maybe putting this include under the ifdef CONFIG_SECURE_FLASH_ENC_ENABLED ?

raffarost (approved):

sylvioalves (approved):

PR #456, opened 2025-06-24
zephyr: snippets: add flash and psram variants snippets

marekmatej:
* Add FLASH related snippets.
* Add PSRAM related snippets.


marekmatej:
Added in https://github.com/zephyrproject-rtos/zephyr/pull/87432

PR #471, opened 2025-08-08
mspi: Zephyr sources for MSPI driver

Pavllick:
Source files added for new MSPI driver to support [PR #94140](https://github.com/zephyrproject-rtos/zephyr/pull/94140) to Zephyr Project.

### DIFF
--- a/zephyr/esp32/CMakeLists.txt
+++ b/zephyr/esp32/CMakeLists.txt
@@ -175,6 +175,13 @@ if(CONFIG_SOC_SERIES_ESP32)
     )
 
   zephyr_sources_ifdef(
+    CONFIG_MSPI_ESP32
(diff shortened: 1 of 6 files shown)


zephyr/esp32/CMakeLists.txt
```diff
@@ -174,6 +174,13 @@ if(CONFIG_SOC_SERIES_ESP32)
     ../../components/soc/lldesc.c
     )
 
+  zephyr_sources_ifdef(
```
marekmatej, line 177: I don't think this is a good way to go. You are adding the same sources again for the "different" peripheral.

marekmatej (changes requested):
I dont think this is a way to go. Will comment on zephyr PR too.

PR #535, opened 2026-03-25
hal: wifi: Add PM support

raffarost:
Add PM support for wifi targets.


sidmodi-mw:
@raffarost Wondering if you are planning on moving this on from draft state to ready state? I am very much looking forward to PM support for WiFi so thank you for implementing this!

components/esp_wifi/src/wifi_init.c
```diff
@@ -128,14 +157,15 @@ static void esp_wifi_set_log_level(void)
...
+#if (CONFIG_PM && SOC_PM_MODEM_RETENTION_BY_REGDMA)
 static esp_err_t init_wifi_mac_sleep_retention(void *arg)
 {
     int config_size;
     sleep_retention_entries_config_t *config = esp_wifi_internal_mac_retention_context_get(&config_size);
     esp_err_t err = sleep_retention_entries_create(config, config_size, 3, SLEEP_RETENTION_MODULE_WIFI_MAC);
-    ESP_RETURN_ON_ERROR(err, TAG, "failed to allocate memory for modem (%s) retention", "WiFi MAC");
-    ESP_LOGD(TAG, "WiFi MAC sleep retention initialization");
+    if (err != ESP_OK) {
+        ESP_LOGW(TAG, "failed to allocate memory for modem (%s) retention", "WiFi MAC");
```
sylvioalves: should it `return err;` here?

components/esp_hw_support/sleep_modem.c
```diff
@@ -35,7 +35,7 @@ static void esp_pm_light_sleep_default_params_config(int min_freq_mhz, int max_f
 #if SOC_PM_RETENTION_HAS_CLOCK_BUG && CONFIG_MAC_BB_PD
 static bool s_modem_sleep = false;
 static uint8_t s_modem_prepare_ref = 0;
-static _lock_t s_modem_prepare_lock;
+static struct k_mutex s_modem_prepare_lock = Z_MUTEX_INITIALIZER(s_modem_prepare_lock);
```
sylvioalves: Why not just `K_MUTEX_DEFINE(s_modem_prepare_lock);` instead?

components/esp_wifi/src/wifi_init.c
```diff
@@ -108,6 +111,32 @@ wifi_mac_time_update_cb_t s_wifi_mac_time_update_cb = NULL;
 
 static const char *TAG = "wifi_init";
 
+#if CONFIG_PM
+static void esp_wifi_pm_policy_state_lock_get(void)
```
sylvioalves, line 115: Can you use esp_os_ critical calls instead? Current implementation will miss on/off states if put/get are called without sync:

```c
#include "esp_private/critical_section.h"

static DEFINE_CRIT_SECTION_LOCK_STATIC(s_pm_lock);
static uint32_t s_pm_policy_refcnt;

 static void esp_wifi_pm_policy_state_lock_get(void)
 {
    esp_os_enter_critical(&s_pm_lock);
    if (s_pm_policy_refcnt++ == 0) {
         pm_policy_state_all_lock_get();
     }
    esp_os_exit_critical(&s_pm_lock);
 }

 static void esp_wifi_pm_policy_state_lock_put(void)
 {
    esp_os_enter_critical(&s_pm_lock);
    if (s_pm_policy_refcnt > 0 && --s_pm_policy_refcnt == 0) {
         pm_policy_state_all_lock_put();
     }
    esp_os_exit_critical(&s_pm_lock);
 }
```

raffarost:
> @raffarost Wondering if you are planning on moving this on from draft state to ready state? I am very much looking forward to PM support for WiFi so thank you for implementing this!

hi @sidmodi-mw. For sure. This PR depends on sleep retention, which depends on other things. So at the moment it has to wait until the pipeline moves, but it will eventually :)

components/esp_wifi/src/wifi_init.c
```diff
@@ -213,32 +233,28 @@ static esp_err_t wifi_deinit_internal(void)
...
-    esp_pm_unregister_skip_light_sleep_callback(esp_wifi_internal_is_tsf_active);
+#if CONFIG_PM && SOC_WIFI_HW_TSF
     esp_pm_unregister_inform_out_light_sleep_overhead_callback(esp_wifi_internal_update_light_sleep_wake_ahead_time);
     esp_sleep_disable_wifi_wakeup();
 # if CONFIG_ESP_WIFI_ENHANCED_LIGHT_SLEEP
     esp_sleep_disable_wifi_beacon_wakeup();
 # endif
-#endif /* SOC_WIFI_HW_TSF */
+#endif /* CONFIG_PM && SOC_WIFI_HW_TSF */
 #if SOC_PM_MODEM_RETENTION_BY_REGDMA
```
sylvioalves: Initing sleep_retentions uses CONFIG_PM and is missing here.
```#if (CONFIG_PM && SOC_PM_MODEM_RETENTION_BY_REGDMA)```

File: components/esp_wifi/src/wifi_init.c [hunk 5]
```diff
@@ -1,11 +1,12 @@
 /*
- * SPDX-FileCopyrightText: 2015-2025 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2015-2026 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
 
 #include <zephyr/kernel.h>
```
sylvioalves, line 7: This file has 4-space formatting style and some changes are using TAB. Would you make default 4-spaces instead?

components/esp_wifi/src/wifi_init.c
sylvioalves: Should you also use CONFIG_PM here and in the free call?

zephyr/port/include/zephyr_compat.h
```diff
@@ -128,6 +132,30 @@
 #define CONFIG_ESP_PHY_MAX_TX_POWER CONFIG_ESP32_PHY_MAX_TX_POWER
 #endif
 
+#ifdef CONFIG_ESP32_WIFI_ENHANCED_LIGHT_SLEEP
+#define CONFIG_ESP_WIFI_ENHANCED_LIGHT_SLEEP CONFIG_ESP32_WIFI_ENHANCED_LIGHT_SLEEP
```
marekmatej, line 140: Since the light sleep in this PR was replaced, do we need this?

marekmatej (commented):
LGTM, with one question

zephyr/port/include/zephyr_compat.h
raffarost: I'm not sure what you mean by replaced :)
But this is a hardware feature that is still present indeed for some chips, and since it can be selected by the user on Zephyr side, we have to keep an alias (current solution at least) to also select in on the HAL side.

marekmatej (approved):

components/esp_wifi/src/wifi_init.c
```diff
@@ -108,6 +111,26 @@ wifi_mac_time_update_cb_t s_wifi_mac_time_update_cb = NULL;
...
+    if (s_pm_lock_refcnt++ == 0) {
+        pm_policy_state_all_lock_get();
+    }
+    esp_os_exit_critical(&s_pm_lock);
+}
+
+static void esp_wifi_pm_policy_state_lock_put(void)
+{
+    esp_os_enter_critical(&s_pm_lock);
+    if (s_pm_lock_refcnt > 0 && --s_pm_lock_refcnt == 0) {
```
sylvioalves, line 127: I think you only need: `if (--s_pm_lock_refcnt == 0) {`, right?

components/esp_wifi/src/wifi_init.c
```diff
@@ -128,22 +151,24 @@ static void esp_wifi_set_log_level(void)
...
+#if (CONFIG_PM && SOC_PM_MODEM_RETENTION_BY_REGDMA)
 static esp_err_t init_wifi_mac_sleep_retention(void *arg)
 {
     int config_size;
     sleep_retention_entries_config_t *config = esp_wifi_internal_mac_retention_context_get(&config_size);
     esp_err_t err = sleep_retention_entries_create(config, config_size, 3, SLEEP_RETENTION_MODULE_WIFI_MAC);
-    ESP_RETURN_ON_ERROR(err, TAG, "failed to allocate memory for modem (%s) retention", "WiFi MAC");
-    ESP_LOGD(TAG, "WiFi MAC sleep retention initialization");
+    if (err != ESP_OK) {
+        ESP_LOGW(TAG, "failed to allocate memory for modem (%s) retention", "WiFi MAC");
```
sylvioalves: Perhaps ESP_LOGE here as it failed?

components/esp_wifi/src/wifi_init.c
```diff
@@ -108,6 +111,26 @@ wifi_mac_time_update_cb_t s_wifi_mac_time_update_cb = NULL;
 
 static const char *TAG = "wifi_init";
 
+#if CONFIG_PM
+static void esp_wifi_pm_policy_state_lock_get(void)
+{
+    esp_os_enter_critical(&s_pm_lock);
+    if (s_pm_lock_refcnt++ == 0) {
+        pm_policy_state_all_lock_get();
```
sylvioalves, line 119: Would you check pm policy here if it should really lock all or just `pm_policy_state_lock_get(PM_STATE_STANDBY, PM_ALL_SUBSTATES);` ? Same for the put call after.

components/esp_wifi/src/wifi_init.c
raffarost: The check for a positive value of `s_pm_lock_refcnt` is just protecting form an async call to `esp_wifi_pm_policy_state_lock_put`, meaning for some reason the ROM trying to do a put when not holding a lock. I'd keep it for safety, even if in ideal conditions it shouldn't happen...

components/esp_wifi/src/wifi_init.c
raffarost: Since `PM_STATE_SOFT_OFF` is disabled in the device tree (it's there for reference only), it's very unlikely that users will use a `pm_state_force()` call to trigger a `deep_sleep` (sys_poweroff() is the usual way). So for ESP devices we actually have one power scheme only that will trigger device suspend actions. That's why a simple lock (non-discriminative) is being used so far. Besides, blocking both schemes while any operation is oingoing seems correct too.

sylvioalves (approved):

wmrsouza (approved):

PR #554, opened 2026-05-12
tools: add find-reachable-from-loader-func cmd to check_callgraph.py

almir-okato:
find-reachable-from-loader-func find the call/object references from a given root function that should be placed in the i/dram_loader_seg of MCUboot bootloader (works for Espressif Port and Zephyr Port).

### DIFF
--- a/tools/ci/check_callgraph.py
+++ b/tools/ci/check_callgraph.py
@@ -7,8 +7,9 @@
 import argparse
 import os
 import re
+from collections import deque
 from functools import partial
-from typing import BinaryIO, Callable, Dict, Generator, List, Optional, Tuple
+from typing import BinaryIO, Callable, Dict, Generator, List, Optional, Tuple, Set, Deque
 
 import elftools
 from elftools.elf import elffile
@@ -60,22 +61,33 @@ TARGET_SECTIONS: Dict[str, List[SectionAddressRange]] = {
 
 
 class Symbol(object):
-    def __init__(self, name: str, addr: int, local: bool, filename: Optional[str], section: Optional[str]) -> None:
+    def __init__(self, name: str, addr: int, local: bool, filename: Optional[str], section: Optional[str], st_type: Optional[str] = None, bind: Optional[str] = None, origin: Optional[str] = None) -> None:
         self.name = name
         self.addr = addr
         self.local = local
         self.filename = filename
         self.section = section
+        self.st_type = st_type
+        self.bind = bind
+        self.origin = origin
         self.refers_to: List[Symbol] = list()
         self.referred_from: List[Symbol] = list()
 
+    def is_func(self) -> bool:
+        return self.st_type == 'STT_FUNC'
+
+    def is_object(self) -> bool:
+        return self.st_type in ('STT_OBJECT',)
+
     def __str__(self) -> str:
-        return '{} @0x{:08x} [{}]{} {}'.format(
+        return '{} @0x{:08x} [{}]{} {}{}{}'.format(
             self.name,
             self.addr,
             self.section or 'unknown',
             ' (local)' if self.local else '',
-            self.filename
+            self.filename or '',
+            f' [{self.st_type}]' if self.st_type else '',
+            f' <{self.origin}>' if self.origin else ''
         )
 
 
@@ -114,11 +126,14 @@ class ElfInfo(object):
                 continue
             filename = None
             for sym in s.iter_symbols():
-                sym_type = sym.entry['st_info']['type']
+                st_info = sym.entry['st_info']
+                sym_type = st_info['type']
+                bind = st_info['bind']
                 if sym_type == 'STT_FILE':
                     filename = sym.name
+                    continue
                 if sym_type in ['STT_NOTYPE', 'STT_FUNC', 'STT_OBJECT']:
-                    local = sym.entry['st_info']['bind'] == 'STB_LOCAL'
+                    local = bind == 'STB_LOCAL'
                     addr = sym.entry['st_value']
                     symbols.append(
                         Symbol(
@@ -127,6 +142,8 @@ class ElfInfo(object):
                             local,
                             filename if local else None,
                             self.section_for_addr(addr),
+                            sym_type,
+                            bind,
                         )
                     )
         return symbols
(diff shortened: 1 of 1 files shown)


tools/ci/check_callgraph.py
```diff
@@ -320,14 +336,127 @@ def match_rtl_funcs_to_symbols(rtl_functions: List[RtlFunction], elfinfo: ElfInf
...
+            return 'DRAM_LOADER'
+    return None
+
+
+def _in_loader_section(sec: Optional[str]) -> bool:
+    return bool(sec) and ('loader' in sec)
+
+
+def _origin_pattern(origin: str) -> str:
+    import os as _os
```
sylvioalves: `os` is already imported at the top of the file. I think you can remove from here and replace `_os` to `os`.

tools/ci/check_callgraph.py
```diff
@@ -320,14 +336,127 @@ def match_rtl_funcs_to_symbols(rtl_functions: List[RtlFunction], elfinfo: ElfInf
...
+    # Build adjacency from existing references
+    adj: Dict[Symbol, List[Symbol]] = {}
+    by_name: Dict[str, List[Symbol]] = {}
+    for s in symbols:
+        by_name.setdefault(s.name, []).append(s)
+    for r in refs:
+        adj.setdefault(r.from_sym, []).append(r.to_sym)
+
+    # Initialize frontier with all symbol objects that match roots
+    sym_queue: List[Symbol] = []
```
sylvioalves: Considering using `sym_queue: Deque[Symbol] = deque()` instead and `cur = sym_queue.popleft()`. not mandatory though.

sylvioalves:
In `_classify_placement`, default DRAM matching only triggers on `.dram*` sections.. does that mean a function/object placed in plain `.rodata.x/.bss.x/.data.x` is intentionally skipped from `DRAM_LOADER` classification unless the caller passes `--dram-sections`? Also, in the `--summary` output, was it intended to print the IRAM linker fragment (.literal .text .iram0 ...) for DRAM_LOADER groups and the BSS/data fragments for IRAM_LOADER groups? Both seem to point the user at the wrong segment.

almir-okato:
> In `_classify_placement`, default DRAM matching only triggers on `.dram*` sections.. does that mean a function/object placed in plain `.rodata.x/.bss.x/.data.x` is intentionally skipped from `DRAM_LOADER` classification unless the caller passes `--dram-sections`? Also, in the `--summary` output, was it intended to print the IRAM linker fragment (.literal .text .iram0 ...) for DRAM_LOADER groups and the BSS/data fragments for IRAM_LOADER groups? Both seem to point the user at the wrong segment.

Checking `sec.startswith( ... )` should cover this kind of case, there are some default name checking.

Regarding the `--summary`, I changed it to print the linker script's placement recommendation of the objects containing the symbols found.

tools/ci/check_callgraph.py
```diff
@@ -320,14 +337,126 @@ def match_rtl_funcs_to_symbols(rtl_functions: List[RtlFunction], elfinfo: ElfInf
...
 
-def get_symbols_and_refs(rtl_list: List[str], elf_file: BinaryIO, ignore_pairs: List[IgnorePair]) -> Tuple[List[Symbol], List[Reference]]:
+def _classify_placement(sym: Symbol, iram_like: List[str], dram_like: List[str]) -> Optional[str]:
+    sec = sym.section or ''
+    # IRAM code placement: any function in text/iram sections
+    if sym.is_func() and any(sec.startswith(s) for s in iram_like + ['.text', '.iram', '.iram.text', '.iram0.text']):
+        return 'IRAM_LOADER'
+    # DRAM data/rodata/bss placement
+    if (sym.is_object() or sec.startswith('.rodata') or sec.startswith('.bss') or '.rodata' in sec or '.bss' in sec or '.data' in sec):
+        if any(sec.startswith(s) for s in dram_like + ['.dram', '.dram0', '.dram0.bss', '.dram0.data', '.dram0.rodata']):
```
sylvioalves: Should you also add here `'.rodata', '.bss', '.data', '.sbss', '.sdata', '.srodata'` as well?  In bootloader.ld files, those areas lives in DRAM area, right?

tools/ci/check_callgraph.py
```diff
@@ -7,8 +7,9 @@
 import argparse
 import os
 import re
+from collections import deque
 from functools import partial
-from typing import BinaryIO, Callable, Dict, Generator, List, Optional, Tuple
+from typing import BinaryIO, Callable, Dict, Generator, List, Optional, Tuple, Set
```
sylvioalves: Missing `Deque` here?

tools/ci/check_callgraph.py
```diff
@@ -320,14 +337,126 @@ def match_rtl_funcs_to_symbols(rtl_functions: List[RtlFunction], elfinfo: ElfInf
...
     for file_name in rtl_list:
         load_rtl_file(file_name, file_name, rtl_functions, ignore_pairs)
 
-    return match_rtl_funcs_to_symbols(rtl_functions, elfinfo)
+    symbols, refs = match_rtl_funcs_to_symbols(rtl_functions, elfinfo)
+
+    # Optionally annotate symbols with origin from map
+    if map_path:
+        symbol_names: Set[str] = {s.name for s in symbols}
+        origin_map = build_origin_map(map_path, symbol_names)
```
sylvioalves: What happens when two same-named static symbols from different object files?  Perhaps it should check name,addr instead?

tools/ci/check_callgraph.py
```diff
@@ -424,6 +573,99 @@ def main() -> None:
...
+                '(.sbss .sbss.* .sbss2 .sbss2.* .bss .bss.*)',
+            ),
+            (
+                '.dram_loader.data',
+                'dram_loader_seg',
+                '(.data .data.* .rodata .rodata.* .sdata .sdata.* .sdata2 .sdata2.* .srodata .srodata.* .dram*)',
+            ),
+        ]
+
+        def print_loader_linker_summary(origins: List[str]) -> None:
```
raffarost, line 733: From AI assisted review:

When the `--summary` path runs, it always emits linker fragments for all three segments regardless of what was actually found. If a user runs this against a binary where only IRAM symbols are out of place, the output will still suggest adding object patterns to .dram_loader.bss and .dram_loader.data — which would be wrong linker script additions. That can actively mislead whoever is acting on the output.

The fix is straightforward: pass two separate origin sets to the summary function (one for IRAM, one for DRAM) and only emit the segment entries that have at least one pattern.

tools/ci/check_callgraph.py
almir-okato: I thought those were already caught. Now I changed it to match the sections prefix similarly as for functions symbols (IRAM).

tools/ci/check_callgraph.py
almir-okato: I got a case like that, yes, name+addr solves that.

tools/ci/check_callgraph.py
almir-okato: I print this way just to ensure the whole file/object will have its symbols in all loader sections.

raffarost (approved):

sylvioalves (approved):

wmrsouza (approved):

marekmatej (approved):

PR #556, opened 2026-05-13
mbedtls: Use private/ includes for TF-PSA-Crypto legacy crypto

krish2718:
Align ESP-IDF port wrappers, WPA supplicant crypto, BLE SM crypto, and ROM OSI shim with TF-PSA-Crypto layout: legacy algorithm headers live under mbedtls/private/. Update the Zephyr mbedtls sync patch accordingly.

Assisted-by: Cursor: Auto

### DIFF
--- a/components/mbedtls/port/include/mbedtls/bignum.h
+++ b/components/mbedtls/port/include/mbedtls/bignum.h
@@ -5,13 +5,8 @@
  */
 #pragma once
 
-/* Zephyr ships mbedtls 3.6.5, which does not expose mbedtls/private/bignum.h.
- * Upstream ESP-IDF uses #include_next "mbedtls/private/bignum.h" with
- * MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS. When Zephyr upgrades to a mbedtls
- * version that provides the private/ namespace, revert this to the upstream
- * form.
- */
-#include_next "mbedtls/bignum.h"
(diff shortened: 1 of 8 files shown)


components/mbedtls/port/mbedtls_rom/mbedtls_rom_osi.h
```diff
@@ -7,34 +7,34 @@
 #pragma once
 
 #include <stddef.h>
-#include "mbedtls/aes.h"
+#include "mbedtls/private/aes.h"
```
krish2718: I guess this file shouln't have been touched as its already ROMed?

sylvioalves: Yes. Also, this file is not touched by Zephyr at all because it is for the mbedtls 3.6.x use only. So I restored both files to original as not needed.

```c
components/mbedtls/port/mbedtls_rom/mbedtls_rom_osi.h
components/bt/porting_btdm/controller/ble/src/ble.c
```

krish2718: I don't see the restore in your push, are you planning a second push?

sylvioalves: Just did.

sylvioalves (approved):

marekmatej (approved):

PR #599, opened 2026-06-20
west: cleanup tools.py and remove deprecated west.log module

pdgendt:
- Copied the `ruff` configuration from Zephyr
- Used `ruff` to cleanup some formatting/linter issues.
- Removed the deprecated `west.log` module

I tried to keep the changes as close to the original code as possible. There are some alternatives; move free functions inside the `WestCommand` class, or pass the command instance. But I'm not sure how the HAL tooling is considered public API.

### DIFF
--- a/west/tools.py
+++ b/west/tools.py
@@ -35,14 +35,6 @@ baud_rate = None
 target_soc = None
 
 
-def cmd_check(cmd, cwd=None, stderr=subprocess.STDOUT):
-    return subprocess.check_output(cmd, cwd=cwd, stderr=stderr)
-
-
-def cmd_exec(cmd, cwd=None, shell=False):
-    return subprocess.check_call(cmd, cwd=cwd, shell=shell)
-
-
 def _normalize_chip(name):
     return name.lower().replace('-', '')
 
@@ -271,4 +263,4 @@ class Tools(WestCommand):
         if not args.enable_address_decoding:
             cmd.append("-d")
 
-        cmd_exec(cmd, cwd=cmd_path)
+        subprocess.check_call(cmd, cwd=cmd_path)
--- a/west/tools.py
+++ b/west/tools.py
@@ -35,6 +35,7 @@ build_elf_path = None
 baud_rate = None
 target_soc = None
 
+
 def cmd_check(cmd, cwd=None, stderr=subprocess.STDOUT):
     return subprocess.check_output(cmd, cwd=cwd, stderr=stderr)
 
@@ -50,6 +51,7 @@ def _normalize_chip(name):
 def get_esp_serial_port(module_path, skip_soc_match=False):
     try:
         import serial.tools.list_ports
+
         esptool_path = os.path.join(module_path, 'tools', 'esptool_py')
         sys.path.insert(0, esptool_path)
         import esptool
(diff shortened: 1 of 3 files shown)


west/tools.py
```diff
@@ -79,60 +85,66 @@ def get_esp_serial_port(module_path, skip_soc_match=False):
...
+            sys.exit(1)
 
         serial_port = esp.serial_port
         esp._port.close()
 
         return serial_port
-    except Exception as e:
-        log.die("Error detecting ESP serial port: {}".format(e))
-        return None
+    except Exception:
```
sylvioalves, line 115: @pdgendt I'll add some comments here. Let me know if you want me to perform those changes or if you do it.

Can you restore the Exception as `e` + logger message?

west/tools.py
```diff
@@ -79,60 +85,66 @@ def get_esp_serial_port(module_path, skip_soc_match=False):
...
     def load_runners_yaml(path):
         # Load runners.yaml and convert to Python object.
 
         try:
-            with open(path, 'r') as f:
+            with open(path) as f:
                 content = yaml.safe_load(f.read())
-        except Exception as e:
-            log.die("runners.yaml file open error: {}".format(e))
+        except Exception:
```
sylvioalves, line 134: same here

west/tools.py
```diff
@@ -35,6 +34,7 @@
 baud_rate = None
 target_soc = None
 
+
 def cmd_check(cmd, cwd=None, stderr=subprocess.STDOUT):
```
sylvioalves: Would you remove `cmd_check`. It is unused.

west/tools.py
pdgendt: This is already done using `_logger.exception`

sylvioalves:
Well, there is no CI/.github workflow for Ruff check (yet). I think this file can stay anyway and we can add a follow up to enable it.

west/tools.py
pdgendt: Will do

sylvioalves:
@pdgendt Hi, I think this change is causing some inconsistencies in the output. See main vs PR:



1) When no USB connected: (2 "ERRORS")

main:
```c
FATAL ERROR: No serial ports found. Connect a device, or use '-p PORT' option to set a specific port.
```

PR:
```c 
ERROR: west.commands.ext.cmd_1: ERROR: No serial ports found. Connect a device, or use '-p PORT' option to set a specific port.`
``` 
--

2) The logger `__name__` itself: `west.commands.ext.cmd_1`. perhaps both `__name__` should be replaced to 'Monitor' instead.

main:
```c
west espressif monitor
Connecting...
Detecting chip type... ESP32-P4
Auto-selected /dev/ttyACM0 for esp32p4
Hard resetting via RTS pin...
``` 

PR:
```c
west espressif monitor
Connecting....
Detecting chip type... ESP32-P4
west.commands.ext.cmd_1: INFO: Auto-selected /dev/ttyACM0 for esp32p4
Hard resetting via RTS pin...
```

pdgendt:
> @pdgendt Hi, I think this change is causing some inconsistencies in the output. See main vs PR:
> 
> 1. When no USB connected: (2 "ERRORS")

This is in upstream Zephyr, will open a PR.

> 2. The logger `__name__` itself: `west.commands.ext.cmd_1`. perhaps both `__name__` should be replaced to 'Monitor' instead.

Used `"esp_monitor"`, this should avoid any collision.

pdgendt:
See https://github.com/zephyrproject-rtos/zephyr/pull/111783

sylvioalves (approved):

tmedicci (approved):

wmrsouza (approved):

PR #613, opened 2026-07-13
hal: pm: Fix sleep retention

raffarost:
Fix sleep retention for several peripherals after a recent HAL sync, which requires some file imports (as is), and refactored xxxx_priv.h header files to keep Zephyr driver usage simple, while stripping OS specific code present in original HAL code.

### DIFF
--- a/tools/sync/hal-config.yml
+++ b/tools/sync/hal-config.yml
@@ -17,7 +17,10 @@ components:
   - esp_coex
   - esp_common
   - esp_driver_dac
+  - esp_driver_dma
   - esp_driver_gpio
+  - esp_driver_gptimer
+  - esp_driver_mcpwm
   - esp_driver_touch_sens
   - esp_driver_tsens
   - esp_event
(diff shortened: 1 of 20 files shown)


tools/sync/hal-config.yml
tmedicci, line 1: Hi @raffarost , I didn't get exactly: did this come from a new HAL sync? If so, shouldn't the `esp_idf_sha` parameter be refreshed? Or did you use the tooling based on the same ESP-IDF commit SHA?

tools/sync/hal-config.yml
```diff
@@ -111,6 +114,8 @@ exclude:
   - esp_common/test_apps
   - esp_driver_dac/test_apps
   - esp_driver_gpio/test_apps
```
sylvioalves, line 117: please add `- esp_driver_dma/test_apps`

tools/sync/hal-config.yml
sylvioalves: keep as is, the new hal-sync will update the sha properly. This is only cherry-picking content from current synced esp-idf.

raffarost: This is just a fix following some changes on sleep retention that the new HAL adopted. We lacked bringing some additional files and adjusting datatypes (essentially), as PM configs are not yet covered by CI.

sylvioalves (approved):

tmedicci (approved):

marekmatej (approved):

PR #615, opened 2026-07-16
zephyr: route hal os calls through the esp_os glue

sylvioalves:
Replace every Wi-Fi, coex, Bluetooth controller, NimBLE porting, esp_timer, supplicant port, heap binding and HAL internal direct Zephyr kernel call with the esp_os glue, so no source references a kernel object directly. Thread stacks are created dynamically, matching how the controllers expect the OS to own task stack memory, so the static stack definitions are removed.

The glue itself lives in the Zephyr tree under
modules/hal_espressif/os_wrapper, alongside the other vendor OS wrappers. Switch the module to an external cmake so that tree-side entry point builds the glue, while the HAL Kconfig keeps the standard module entry to avoid a recursive source at repo root.

### DIFF
--- a/components/bt/controller/esp32/bt.c
+++ b/components/bt/controller/esp32/bt.c
@@ -7,6 +7,7 @@
 #include <zephyr/kernel.h>
 #include <zephyr/sys/printk.h>
 #include <zephyr/random/random.h>
+#include <esp_os.h>
 
 #include <stddef.h>
 #include <stdlib.h>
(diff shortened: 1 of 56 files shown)


tmedicci (approved):

wmrsouza (approved):

marekmatej (approved):
needs rebase

raffarost (approved):

PR #617, opened 2026-07-21
wifi: fix wpa2 tkip crash caused by rom hmac_md5

sylvioalves:
Stop linking the ROM hmac_md5 (and other syscall-table dependent libc exports), whose internal ROM malloc dereferences the never installed syscall_table_ptr, crashing TKIP connections on ESP32-S3/C3. Also fix os_get_time so supplicant eloop timers (EAPOL retransmissions) actually fire


marekmatej (approved):
nice catches!

raffarost (approved):

wmrsouza (approved):

PR #619, opened 2026-07-27
esp32s3/p4: add smp support

sylvioalves:
Add the changes hal_espressif needs to support dual-core SMP on the
ESP32-S3 and ESP32-P4.

Locking primitives are made SMP safe so they exclude the other core
rather than only local interrupts.

The Wi-Fi, Bluetooth and esp_timer tasks are pinned to a fixed core,
since the radio blobs assume their task and interrupt share a core and
the timer task stays with the stacks it serves. Each is created
suspended, pinned, then started.

The Wi-Fi task affinity choice is dropped here because it is now
defined on the Zephyr side, where it also selects the core the Wi-Fi
interrupt is reserved on.

The Wi-Fi blob's interrupt lines (WMAC and WBB) are reserved on both
cores. The interrupt allocator reserves per core, so reserving them on
one core only let the other core's allocator hand the same lines to
the scheduler IPI and the system timer, stalling the secondary core.

### DIFF
--- a/components/bt/controller/esp32/bt.c
+++ b/components/bt/controller/esp32/bt.c
@@ -833,11 +833,23 @@ static int32_t IRAM_ATTR queue_recv_from_isr_wrapper(void *queue, void *item, vo
 
 static int32_t task_create_wrapper(void *task_func, const char *name, uint32_t stack_depth, void *param, uint32_t prio, void *task_handle, uint32_t core_id)
 {
+    ARG_UNUSED(core_id);
+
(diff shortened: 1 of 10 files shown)


tmedicci (approved):

sylvioalves:
@marekmatej @raffarost PTAL

sylvioalves:
@raffarost @marekmatej PTAL, this is ready and tested.

sylvioalves:
Closing so I split PRs.

PR #620, opened 2026-07-29
fix: enforce minimum 64-byte temp buffer for short PSRAM flash reads

### LINKED ISSUE #114822: esp32: flash read to PSRAM buffer fails when length is below 64 bytes
### Describe the bug

On ESP32-S3 with SPIRAM enabled, flash reads fail with -EIO when the destination buffer is in PSRAM and the requested length is below 64 bytes.

This breaks ZMS/settings operations performed from a PSRAM-backed thread stack. In my case, storing Wi-Fi credentials always failed.

The problem is in get_buffer_malloc() in spi_flash_os_func_app.c. read_chunk_size is smaller than 64 for short reads, so this loop never runs:

while (ret == NULL && read_chunk_size >= 64)

temp_buffer remains NULL and esp_flash_read() returns ESP_ERR_NO_MEM even when heap memory is available.

The issue became visible after Zephyr commit 31660694775d1b1b5916dbdca1f94e892e334f66 enabled the OS-aware flash functions. The faulty allocation loop was introduced by hal_espressif commit f497c7d00b46cd4597f36559e2a3d566004dc9c3.

### Regression

- [x] This is a regression.

### Steps to reproduce

1. Build for esp32s3_devkitc/esp32s3/procpu with CONFIG_FLASH=y and CONFIG_ESP_SPIRAM=y.
2. Call flash_read() after boot with a PSRAM-backed destination buffer and a length below 64 bytes.
3. flash_read() returns -EIO and the driver logs error 257.

It can also be reproduced by saving a setting through ZMS from a thread whose stack is in PSRAM.

### Relevant log output

```shell
flash_esp32: Flash read error: 257
flash_esp32: Flash read error: -5
wifi_credentials: Failed to store WiFi credentials at index 0, err: -5
```

### Impact

Functional Limitation – Some features not working as expected, but system usable.

### Environment

- OS: Debian GNU/Linux 13
- Board: esp32s3_devkitc/esp32s3/procpu
- Toolchain: Zephyr SDK 1.0.1
- Zephyr: af17c0c8aa2a82a0b714a423915ad3610063ee68
- hal_espressif: 150e0c1bd24fd568864ae50fa690e9ec526022a4

### Additional Context

This one-line change fixes the issue:

size_t read_chunk_size = MAX((reqest_size + 3) & ~3, 64);

After applying it, Wi-Fi credentials can be stored in ZMS and loaded successfully after reboot.

trinityman-hash:
get_buffer_malloc() computed read_chunk_size from the requested read length, so requests below 64 bytes produced a chunk size under 64. The allocation loop only runs while read_chunk_size >= 64, so it never executed, temp_buffer stayed NULL, and esp_flash_read() failed with ESP_ERR_NO_MEM even though heap memory was available.

This broke flash reads into PSRAM-backed buffers under 64 bytes, including ZMS/settings writes from PSRAM-backed thread stacks (e.g. storing Wi-Fi credentials).

Fix: clamp read_chunk_size to a 64-byte floor before entering the loop, so short requests still get an eligible temp buffer.

Fixes zephyrproject-rtos/zephyr#114822

### DIFF
--- a/components/spi_flash/spi_flash_os_func_app.c
+++ b/components/spi_flash/spi_flash_os_func_app.c
@@ -219,7 +219,7 @@ static void* get_buffer_malloc(void* arg, size_t reqest_size, size_t* out_size)
      * use exponential backoff down to a minimum of 64 bytes.
      */
     void* ret = NULL;
-    size_t read_chunk_size = (reqest_size + 3) & ~3;
+    size_t read_chunk_size = MAX((reqest_size + 3) & ~3, 64);
 
     while (ret == NULL && read_chunk_size >= 64) {
         ret = k_malloc(read_chunk_size);


sylvioalves:
@trinityman-hash thanks!

sylvioalves (approved):

raffarost (approved):

components/spi_flash/spi_flash_os_func_app.c
```diff
@@ -219,7 +219,7 @@ static void* get_buffer_malloc(void* arg, size_t reqest_size, size_t* out_size)
      * use exponential backoff down to a minimum of 64 bytes.
      */
     void* ret = NULL;
-    size_t read_chunk_size = (reqest_size + 3) & ~3;
+    size_t read_chunk_size = MAX((reqest_size + 3) & ~3, 64);
```
tmedicci, line 222: Do you mind fixing this typo `reqest` -> `request`?

marekmatej (approved):

components/spi_flash/spi_flash_os_func_app.c
sylvioalves: @trinityman-hash would you update the PR fixing this?

sylvioalves:
@tmedicci We can fix the `reqest` typo in next hal-sync (or in esp-idf directly).

PR #622, opened 2026-07-30
zephyr: add Espressif hardware flash encryption support to MCUboot

almir-okato:
Add boot_go_hook to prepare/finalize security enabling routines and encrypts DT flash partitions on first boot. 
Include Kconfigs for hal_espressif configuration and ifdefs compatibility.

Also drop the esp_rom_flash_* wrappers and switch back to bootloader_flash_*. Include the bootloader_flash.c original source under NON_OS_BUILD so SimpleBoot and MCUboot use the ROM/MMU backend.


zephyr/port/bootloader/bootloader_flash.c
```diff
@@ -3,857 +3,11 @@
...
-        spi_mode = ESP_ROM_SPIFLASH_DOUT_MODE;
-    } else if (spi_ctrl & SPI_MEM_FASTRD_MODE) {
-        spi_mode = ESP_ROM_SPIFLASH_FASTRD_MODE;
-    } else {
-        spi_mode = ESP_ROM_SPIFLASH_SLOWRD_MODE;
-    }
-#endif
-    return spi_mode;
-}
+#define NON_OS_BUILD 1
```
marekmatej, line 13: Why cant we use the `NON_OS_BUILD=1` on MCUboot cmake level?

zephyr/esp32/CMakeLists.txt
```diff
@@ -431,6 +430,12 @@ if(CONFIG_SOC_SERIES_ESP32)
     src/soc_random.c
   )
 
+  if (NOT CONFIG_MCUBOOT)
```
marekmatej, line 404: Would it be possible to avoid such single-file asymmetry? Although I remember we moving this before for some reasons.

zephyr/port/boot/mcuboot_flash_encrypt.c
```diff
@@ -0,0 +1,495 @@
...
+	REG_WRITE(SENSITIVE_XTS_AES_KEY_UPDATE_REG, 1);
+#endif
+
+	err = encrypt_bootloader();
+	if (err != ESP_OK && err != ESP_ERR_NOT_FOUND) {
+		return err;
+	}
+
+	/* Encrypt remaining DT flash partitions (slots, scratch, storage, …). */
+	DT_FOREACH_STATUS_OKAY(zephyr_mapped_partition, ENCRYPT_DT_PARTITION);
```
marekmatej, line 332: It would be nice to have a flag marking a partition to encrypt in DTS. But not a blocker.

zephyr/port/boot/mcuboot_flash_encrypt.c
```diff
@@ -0,0 +1,495 @@
...
+#define CRYPT_CNT ESP_EFUSE_SPI_BOOT_CRYPT_CNT
+#define WR_DIS_CRYPT_CNT ESP_EFUSE_WR_DIS_SPI_BOOT_CRYPT_CNT
+#endif
+
+#define FLASH_ENC_CNT_MAX (CRYPT_CNT[0]->bit_count)
+
+#define BOOTLOADER_ADDR   DT_REG_ADDR(DT_NODELABEL(boot_partition))
+#define BOOTLOADER_SIZE   DT_REG_SIZE(DT_NODELABEL(boot_partition))
+#define SLOT0_ADDR        DT_REG_ADDR(DT_NODELABEL(slot0_partition))
+#define SLOT0_SIZE        DT_REG_SIZE(DT_NODELABEL(slot0_partition))
```
marekmatej: Those two macros seem to be unused.

zephyr/port/bootloader/bootloader_flash.c
almir-okato: Because of SimpleBoot, as it needs both flash ROM calls for booting and there are other code points that branches from this macro not being defined.
Before I did some separation on the CMakeLists.txt sources for MCUboot bootloader/app and SimpleBoot, there were also some files conflicting when building.

zephyr/esp32/CMakeLists.txt
almir-okato: I did some separation on the sources for MCUboot bootloader/app and SimpleBoot to avoid compiling unneeded code (also code that may have issues when the bootloader is being built).

zephyr/port/boot/mcuboot_flash_encrypt.c
almir-okato: Yes, this would be a nice feature. I'm sticking to the scope for now.

zephyr/Kconfig.mcuboot
```diff
@@ -0,0 +1,67 @@
...
+	bool "Require flash encryption already enabled in eFuse"
+	help
+	  Refuse to boot if flash encryption is not already enabled.
+
+config SECURE_BOOT_FLASH_ENC_KEYS_BURN_TOGETHER
+	bool "Burn Secure Boot and Flash Encryption keys together"
+	default y if SOC_EFUSE_CONSISTS_OF_ONE_KEY_BLOCK
+	help
+	  On SoCs with a single eFuse key block, Secure Boot and Flash
+	  Encryption keys must be burned together.
```
tmedicci, line 63: On ESP-IDF this is a hidden confi because it's not optional for devices that have a single efuse key block. Wouldn't it be better to keep the same here? (Otherwise, users could disable it, and that'd be catastrophic, wouldn't it?)

zephyr/esp32/src/soc_init.c
```diff
@@ -167,3 +167,18 @@ void bootloader_clock_configure(void)
...
+    if (cfg.enable == 1 && cfg.tieh == RTC_VDDSDIO_TIEH_1_8V) {    // VDDSDIO regulator is enabled @ 1.8V
+        cfg.drefh = 3;
+        cfg.drefm = 3;
+        cfg.drefl = 3;
+        cfg.force = 1;
+        rtc_vddsdio_set_config(cfg);
+        esp_rom_delay_us(10); // wait for regulator to become stable
+    }
+#endif // CONFIG_BOOTLOADER_VDDSDIO_BOOST
+}
```
tmedicci, line 184: Can't we just compile it from `components/bootloader_support/src/bootloader_common.c` to avoid the duplication?

almir-okato: Most of `components/bootloader_support/src/bootloader_common.c` functions is unused and it also raise other compilation problems when included. Since this is only called by `esp32` when it has PSRAM enabled, I rather copied the function.

zephyr/Kconfig.mcuboot
almir-okato: Fixed that, thank you!

zephyr/esp32p4/CMakeLists.txt
```diff
@@ -382,85 +383,91 @@ if(CONFIG_SOC_SERIES_ESP32P4)
...
+      ../../components/esp_hw_support/sleep_cache.c
+      ../../components/esp_hw_support/sleep_retention.c
+      ../../components/esp_hw_support/regi2c_ctrl.c
+      ../../components/esp_mm/esp_cache_msync.c
+      ../../components/esp_rom/patches/esp_rom_crc.c
+      ../../components/esp_rom/patches/esp_rom_efuse.c
+      ../../components/esp_rom/patches/esp_rom_gpio.c
+      ../../components/esp_rom/patches/esp_rom_clic.c
+      ../../components/esp_rom/patches/esp_rom_print.c
+      ../../components/esp_rom/patches/esp_rom_serial_output.c
```
sylvioalves, line 453: `console_init.c` calls `esp_rom_output_switch_buffer` in MCUboot builds, so `esp_rom_serial_output.c` cannot be app-only except on esp32 and s2. please check.

zephyr/port/bootloader/bootloader_flash.c
```diff
@@ -3,859 +3,13 @@
...
-    } else if (spi_ctrl & SPI_MEM_FREAD_DIO) {
-        spi_mode = ESP_ROM_SPIFLASH_DIO_MODE;
-    } else if (spi_ctrl & SPI_MEM_FREAD_DUAL) {
-        spi_mode = ESP_ROM_SPIFLASH_DOUT_MODE;
-    } else if (spi_ctrl & SPI_MEM_FASTRD_MODE) {
-        spi_mode = ESP_ROM_SPIFLASH_FASTRD_MODE;
-    } else {
-        spi_mode = ESP_ROM_SPIFLASH_SLOWRD_MODE;
-    }
+#ifndef CONFIG_BOOTLOADER_MCUBOOT
```
sylvioalves, line 12: Simple Boot apps call `bootloader_mmap` from `soc_late_init_hook`, so the ROM backend unmaps the running image. Should this be MCUboot only?

```suggestion
#ifdef CONFIG_MCUBOOT
```

sylvioalves:
`soc_random_enable` on c5/p4 never calls `rng_ll_enable()`, so `rng_sample_enable` stays 0 while the key is generated. please add if needed.

zephyr/Kconfig.mcuboot
```diff
@@ -0,0 +1,67 @@
...
+endchoice
+
+config SECURE_FLASH_REQUIRE_ALREADY_ENABLED
+	bool "Require flash encryption already enabled in eFuse"
+	help
+	  Refuse to boot if flash encryption is not already enabled.
+
+config SECURE_BOOT_FLASH_ENC_KEYS_BURN_TOGETHER
+	bool
+	default y if SOC_EFUSE_CONSISTS_OF_ONE_KEY_BLOCK
```
sylvioalves, line 60: without the SECURE_BOOT dependency this XORs true on every later boot and the hook fails

```suggestion
	default y if SOC_EFUSE_CONSISTS_OF_ONE_KEY_BLOCK && SECURE_BOOT_V2_ENABLED
```

zephyr/Kconfig.mcuboot
```diff
@@ -0,0 +1,67 @@
...
+if MCUBOOT
+
+config SECURE_FLASH_ENC_ENABLED
+	bool "Espressif hardware flash encryption in MCUboot"
+	help
+	  Enable hardware-assisted flash encryption in the MCUboot bootloader.
+	  Typically set via sysbuild (SB_CONFIG_MCUBOOT_ESP_FLASH_ENCRYPTION).
+
+if SECURE_FLASH_ENC_ENABLED
+
```
sylvioalves, line 13: sounds development mode still burns `DIS_PAD_JTAG` and `DIS_USB_JTAG` permanently, with no way to opt out, perhaps adding a kconfig here solves it? `flash_encryption_secure_features.c` burns both unless `CONFIG_SECURE_BOOT_ALLOW_JTAG` is set 

```c
config SECURE_BOOT_ALLOW_JTAG
	bool "Allow JTAG debugging (NOT SECURE)"
	depends on SECURE_FLASH_ENCRYPTION_MODE_DEVELOPMENT
	help
	  Skip burning the JTAG disable eFuses on first boot.
```

zephyr/port/boot/mcuboot_flash_encrypt.c
```diff
@@ -0,0 +1,529 @@
...
+
+	/* boot_partition is handled specially above. */
+	if (addr == BOOTLOADER_ADDR) {
+		return ESP_OK;
+	}
+
+	if (size == 0) {
+		return ESP_OK;
+	}
+
```
sylvioalves, line 296: perhaps checking alignment as well?

zephyr/port/boot/mcuboot_flash_encrypt.c
```diff
@@ -0,0 +1,529 @@
...
+	REG_WRITE(SENSITIVE_XTS_AES_KEY_UPDATE_REG, 1);
+#endif
+
+	err = encrypt_bootloader();
+	if (err != ESP_OK && err != ESP_ERR_NOT_FOUND) {
+		return err;
+	}
+
+	/* Encrypt remaining DT flash partitions (slots, scratch, storage, …). */
+	DT_FOREACH_STATUS_OKAY(zephyr_mapped_partition, ENCRYPT_DT_PARTITION);
```
sylvioalves, line 332: iterating boot_partition's siblings would keep this on the main flash, Marek's DT-flag suggestion would solve the same thing more explicitly.

```suggestion
		DT_FOREACH_CHILD_STATUS_OKAY(DT_PARENT(DT_NODELABEL(boot_partition)),
				     ENCRYPT_DT_PARTITION);
```

zephyr/port/include/zephyr_compat.h
```diff
@@ -307,4 +315,41 @@
...
+
+#ifdef CONFIG_SECURE_BOOT_FLASH_ENC_KEYS_BURN_TOGETHER
+#define CONFIG_SECURE_BOOT_FLASH_ENC_KEYS_BURN_TOGETHER 1
+#endif
+
+#ifdef CONFIG_SECURE_FLASH_ENCRYPTION_MODE_DEVELOPMENT
+#define CONFIG_SECURE_FLASH_UART_BOOTLOADER_ALLOW_ENC 1
+#define CONFIG_SECURE_FLASH_UART_BOOTLOADER_ALLOW_CACHE 1
+#endif
+
```
sylvioalves, line 354: esp-idf defaults these to n even in development mode, isn't it?

zephyr/esp32p4/CMakeLists.txt
almir-okato: Fixed, thanks!

zephyr/port/bootloader/bootloader_flash.c
almir-okato: As the `bootloader_flash.c` wrapper was dropped, Simple Boot is also using the ROM backend, so this code path applies for both MCUboot and Simple Boot.
I've adjusted `lp_core_image_init` to call `spi_flash_mmap` instead of `bootloader_mmap`.

almir-okato:
> `soc_random_enable` on c5/p4 never calls `rng_ll_enable()`, so `rng_sample_enable` stays 0 while the key is generated. please add if needed.

Yes, `rng_ll_enable` was missing. 
I changed the `soc_random_enable/disable` wrappers back to `bootloader_random_<soc>` functions from hal_espressif, since they were practically identical with exception of that missing call.

zephyr/Kconfig.mcuboot
almir-okato: Tiago also raised concern about this config, I'm removing it until I add Secure Boot.

zephyr/Kconfig.mcuboot
almir-okato: There's a bit of "mixing" on IDF's Kconfigs prefix naming for "SECURE_BOOT_" and "SECURE_FLASH_", so some configs I've mirrored from there were missing. I added the "Potentially insecure options" submenu for Flash Encryption now.

zephyr/port/boot/mcuboot_flash_encrypt.c
almir-okato: Added.

zephyr/port/boot/mcuboot_flash_encrypt.c
almir-okato: Added your suggestion, thanks! I think that DT-flag may be an improvement for the future.

zephyr/port/include/zephyr_compat.h
almir-okato: I added the "Potentially insecure options" submenu for Flash Encryption from IDF now.

PR #624, opened 2026-07-31
esp_supplicant: add STA PMKSA cache export and staging API

siddhantmodi:
### Summary

- Adds APIs to export, stage, and clear a station PMKSA entry, allowing applications to retain enterprise Wi-Fi session state across resets.
- Staged entries are installed once the target BSSID and AKM are known. Lifetimes continue aging while staged, and cached entries are validated before reuse.
- The application owns secure persistence and must account for elapsed time before staging.

### Follow-up

- This provides the ESP32 implementation needed for a subsequent Zephyr PR that will expose persistent PMKSA caching through the common Wi-Fi management layer and integrate supporting drivers.

### Testing

- Validated PMKSA export and reuse across ESP32-C6 resets with enterprise Wi-Fi. Zephyr ESP32-C6 compilation passes.

### Current limitations

- `export()` could theoretically race with internal PMKSA removal since these APIs are not synchronized with the WiFi task.
  - `export()` obtains a pointer to the current PMKSA, but that entry is owned by the supplicant task. A mutex inside this module would not solve the pointer race because native PMKSA code would never acquire that mutex.
- Only 802.1X and 802.1X-SHA256 AKMs with 32-byte PMKs are supported.
- **Only one PMKSA record can be staged at a time, and it is bound to its BSSID, station address, and AKM.**
  - A future PR will add support for multiple record storage and bring the `esp_supplicant` up to parity with `hostap`.
- Imported lifetimes are capped at 12 hours. Applications must subtract time elapsed across reset before staging.
- The exported structure contains secret PMK material and is not a stable serialized storage format. Secure storage and wiping remain application responsibilities.

### DIFF
--- a/components/wpa_supplicant/esp_supplicant/include/esp_wifi_sta_pmksa_cache.h
+++ b/components/wpa_supplicant/esp_supplicant/include/esp_wifi_sta_pmksa_cache.h
@@ -1,0 +1,91 @@
+/*
+ * SPDX-FileCopyrightText: 2026 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef ESP_WIFI_STA_PMKSA_CACHE_H
+#define ESP_WIFI_STA_PMKSA_CACHE_H
+
+#include <stdint.h>
+
+#include "esp_err.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define ESP_WIFI_STA_PMKSA_MAC_LEN   6U
+#define ESP_WIFI_STA_PMKSA_PMKID_LEN 16U
+#define ESP_WIFI_STA_PMKSA_PMK_LEN   32U
+
+/* Policy limit for imported PMKSA lifetimes. */
+#define ESP_WIFI_STA_PMKSA_MAX_LIFETIME_S 43200U
+
+#define ESP_WIFI_STA_PMKSA_AKM_802_1X        0x000FAC01U /* 00-0F-AC:1 */
+#define ESP_WIFI_STA_PMKSA_AKM_802_1X_SHA256 0x000FAC05U /* 00-0F-AC:5 */
+
+/**
+ * @brief RSN PMKSA record.
+ *
+ * @a pmk is secret. Protect persisted records and wipe caller-owned copies.
+ * This structure is not a stable serialized storage format.
+ */
+typedef struct {
+    uint8_t bssid[ESP_WIFI_STA_PMKSA_MAC_LEN];    /**< Authenticator address. */
+    uint8_t sta_addr[ESP_WIFI_STA_PMKSA_MAC_LEN]; /**< Station address. */
+    uint8_t pmkid[ESP_WIFI_STA_PMKSA_PMKID_LEN];
+    uint8_t pmk[ESP_WIFI_STA_PMKSA_PMK_LEN];
+    uint8_t pmk_len;                 /**< Must be ESP_WIFI_STA_PMKSA_PMK_LEN. */
+    uint32_t akm_suite;              /**< IEEE 802.11 AKM suite selector. */
+    uint32_t expiration_remaining_s; /**< Non-zero, at most the max above. */
+    uint32_t reauth_remaining_s;     /**< At most expiration_remaining_s. */
+} esp_wifi_sta_pmksa_cache_entry_t;
+
+/*
+ * Not synchronized against the Wi-Fi task: stage before connecting, export
+ * while associated, clear while disconnected.
+ */
+
+/**
+ * @brief Export the PMKSA in use, lifetimes relative to now and capped.
+ *
+ * @param[out] entry Always zeroed; populated only on success.
+ * @return ESP_OK, ESP_ERR_INVALID_ARG, ESP_ERR_NOT_FOUND, ESP_ERR_NOT_SUPPORTED
+ *         for an out-of-scope AKM or PMK length, or ESP_ERR_INVALID_STATE.
+ */
+esp_err_t esp_wifi_sta_pmksa_cache_export(esp_wifi_sta_pmksa_cache_entry_t *entry);
+
+/**
+ * @brief Stage a record for the next association to its BSSID.
+ *
+ * Discarded if the station address or AKM mismatch. A non-expired record is
+ * kept if another BSSID is attempted. Replaces any previous record and evicts
+ * the entry it installed.
+ * The caller must deduct time elapsed while the record was stored from the
+ * remaining lifetimes before staging. A matching-BSSID attempt consumes the
+ * staged record except when entry allocation fails, which leaves it for retry.
+ *
+ * @param[in] entry Record to stage.
+ * @return ESP_OK or ESP_ERR_INVALID_ARG if the record is malformed.
+ */
+esp_err_t esp_wifi_sta_pmksa_cache_stage(const esp_wifi_sta_pmksa_cache_entry_t *entry);
+
+/** @brief Discard the staged record and flush every station PMKSA. @return ESP_OK */
+esp_err_t esp_wifi_sta_pmksa_cache_clear(void);
+
+/**
+ * @brief Discard staged and externally installed station PMKSA state.
+ *
+ * Supplicant-learned entries are preserved. An externally installed entry that
+ * is in use is also preserved; call this again once disconnected to remove it.
+ *
+ * @return ESP_OK
+ */
+esp_err_t esp_wifi_sta_pmksa_cache_clear_external(void);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* ESP_WIFI_STA_PMKSA_CACHE_H */
--- a/components/wpa_supplicant/esp_supplicant/src/esp_wifi_sta_pmksa_cache.c
+++ b/components/wpa_supplicant/esp_supplicant/src/esp_wifi_sta_pmksa_cache.c
@@ -1,0 +1,328 @@
+/*
+ * SPDX-FileCopyrightText: 2026 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "utils/includes.h"
+#include "utils/common.h"
+#include "common/defs.h"
+
+#include "esp_wifi_sta_pmksa_cache.h"
+#include "esp_wifi_sta_pmksa_cache_i.h"
+#include "rsn_supp/wpa.h"
+#include "rsn_supp/wpa_i.h"
+#include "rsn_supp/pmksa_cache.h"
+
+/* Record awaiting installation on the next association attempt. */
+static esp_wifi_sta_pmksa_cache_entry_t s_staged;
+static bool s_staged_valid;
+static os_time_t s_staged_expiration;
+static os_time_t s_staged_reauth_time;
+
+/* Lets a retry of the same association reuse the entry already installed. */
+static struct {
+    u8 bssid[ETH_ALEN];
+    u8 spa[ETH_ALEN];
+    u8 pmkid[PMKID_LEN];
+    unsigned int akmp;
+    bool valid;
+} s_installed;
+
+static void pmksa_staged_clear(void)
+{
+    forced_memzero(&s_staged, sizeof(s_staged));
+    s_staged_valid = false;
+    s_staged_expiration = 0;
+    s_staged_reauth_time = 0;
+}
+
+/*
+ * Drop the entry installed by this module, if it is still cached. Returns false
+ * if that entry is in use, leaving it installed and still tracked.
+ */
+static bool pmksa_installed_try_remove(void)
+{
+    struct wpa_sm *sm = &gWpaSm;
+    struct rsn_pmksa_cache_entry *entry;
+
+    if (s_installed.valid && sm->pmksa != NULL) {
+        /* pmksa_cache_get() cannot match on AKM, so check it here. */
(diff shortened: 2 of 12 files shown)


sylvioalves:
@siddhantmodi will check soon.

components/wpa_supplicant/esp_supplicant/src/esp_wifi_sta_pmksa_cache.c
```diff
@@ -0,0 +1,306 @@
...
+{
+    struct wpa_sm *sm = &gWpaSm;
+    struct rsn_pmksa_cache_entry *entry;
+
+    if (s_installed.valid && sm->pmksa != NULL) {
+        /* pmksa_cache_get() cannot match on AKM, so check it here. */
+        entry = pmksa_cache_get(sm->pmksa, s_installed.bssid, s_installed.spa,
+                                s_installed.pmkid, NULL);
+        if (entry != NULL && entry->akmp == (int)s_installed.akmp) {
+            pmksa_cache_remove(sm->pmksa, entry);
```
sylvioalves, line 57: i think you could also apply the same rules as done in `install()` call here before `pmksa_cache_remove(sm->pmksa, entry);`.

```c
            if (pmksa_cache_get_current(sm) == entry) {
                pmksa_cache_clear_current(sm);
            }
```

components/wpa_supplicant/esp_supplicant/src/esp_wifi_sta_pmksa_cache.c
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    if (s_staged_expiration <= now.sec) {
+        pmksa_staged_clear();
+        return false;
+    }
+
+    entry = os_zalloc(sizeof(*entry));
+    if (entry == NULL) {
+        pmksa_staged_clear();
```
sylvioalves: if this fails, it calls pmksa_staged_clear() and permanently discards the app's staged record, so one OOM costs full EAP on every later attempt, right? Should transient failures keep the record staged, the way a BSSID mismatch already does (meaning to remove the `pmksa_staged_clear();`)?

sylvioalves: well, `os_zalloc` failing is a scenario where other things would be failing 1st anyway

components/wpa_supplicant/esp_supplicant/include/esp_wifi_sta_pmksa_cache.h
```diff
@@ -0,0 +1,80 @@
...
+ * while associated, clear while disconnected.
+ */
+
+/**
+ * @brief Export the PMKSA in use, lifetimes relative to now and capped.
+ *
+ * @param[out] entry Always zeroed; populated only on success.
+ * @return ESP_OK, ESP_ERR_INVALID_ARG, ESP_ERR_NOT_FOUND, ESP_ERR_NOT_SUPPORTED
+ *         for an out-of-scope AKM or PMK length, ESP_ERR_INVALID_STATE, or
+ *         ESP_FAIL if the monotonic clock cannot be read.
```
sylvioalves: please fix the `ESP_FAIL` in both `export()` and `stage()` calls as it can't occur.

components/wpa_supplicant/esp_supplicant/src/esp_wifi_sta_pmksa_cache.c
```diff
@@ -0,0 +1,310 @@
...
+
+    if (s_installed.valid && sm->pmksa != NULL) {
+        /* pmksa_cache_get() cannot match on AKM, so check it here. */
+        entry = pmksa_cache_get(sm->pmksa, s_installed.bssid, s_installed.spa,
+                                s_installed.pmkid, NULL);
+        if (entry != NULL && entry->akmp == (int)s_installed.akmp) {
+            if (pmksa_cache_get_current(sm) == entry) {
+                pmksa_cache_clear_current(sm);
+            }
+            pmksa_cache_remove(sm->pmksa, entry);
```
sylvioalves, line 57: @siddhantmodi, it seems this remove a live associated pmksa. Check if the change below makes sense.

```suggestion
            if (pmksa_cache_get_current(sm) != entry) {
                pmksa_cache_remove(sm->pmksa, entry);
            }
```

sylvioalves: I previously suggested you to clear, which made sense, but I think check shoud be to remove only when `get_current` isn't the same entry only.

sylvioalves:
btw, you closed the Zephyr PR. Do you intent to continue supporting this?

components/wpa_supplicant/esp_supplicant/src/esp_wifi_sta_pmksa_cache.c
```diff
@@ -0,0 +1,310 @@
...
+    if (!s_staged_valid) {
+        return false;
+    }
+
+    os_get_reltime(&now);
+
+    if (s_staged_expiration <= now.sec) {
+        pmksa_staged_clear();
+        return false;
+    }
```
sylvioalves, line 267: maybe this can be evaluate after L252 so it does not run on every BSSID association?

sidmodi-mw:
Hey @sylvioalves, I fully intend to keep supporting this. I decided to break down things on the Zephyr side into smaller PRs, starting with an [API-only PR](https://github.com/zephyrproject-rtos/zephyr/pull/115470) first, so that I could get feedback on the design before adding the backends (`hostap` and `esp32`).

I will address your comments today. Just a note that I am going on vacation for 4 weeks starting August 17 so there is a chance we might not be able fully close this PR before that. In any case, I will pick it up when I'm back if this doesn't close. It is a feature that we are planning to ship in a product since it significantly helps reduce the power consumption.

sylvioalves:
> Hey @sylvioalves, I fully intend to keep supporting this. I decided to break down things on the Zephyr side into smaller PRs, starting with an [API-only PR](https://github.com/zephyrproject-rtos/zephyr/pull/115470) first, so that I could get feedback on the design before adding the backends (`hostap` and `esp32`).
> 
> I will address your comments today. Just a note that I am going on vacation for 4 weeks starting August 17 so there is a chance we might not be able fully close this PR before that. In any case, I will pick it up when I'm back if this doesn't close. It is a feature that we are planning to ship in a product since it significantly helps reduce the power consumption.

got it, thanks for the heads up.

components/wpa_supplicant/esp_supplicant/src/esp_wifi_sta_pmksa_cache.c
siddhantmodi: I recommend keeping this as it is because with the current implementation, an expired staged record gets wiped at the next association of any kind. If we move this to after L252, a dead PMK will just sit in s_staged until its own BSSID is attempted or the app calls clear.

Since limiting how long a PMK exists in RAM is the point of the expiry check, moving it later slightly makes that a bit weaker.

Do you agree?

siddhantmodi:
@sylvioalves One more note - as mentioned in the PR body, this implementation supports staging of only one PMKSA record at a time. Once I am back from vacation, I plan to add support for multiple PMKSA caching so that it is more in line with how `hostap` works.

sylvioalves:
would you also check whether we should set the expiration in? Otherwise flush or removal won't make expiry timer to re-arm again. 
https://github.com/zephyrproject-rtos/hal_espressif/blob/main/components/wpa_supplicant/src/rsn_supp/pmksa_cache.c#L306
https://github.com/zephyrproject-rtos/hal_espressif/blob/main/components/wpa_supplicant/src/rsn_supp/pmksa_cache.c#L80

components/wpa_supplicant/esp_supplicant/src/esp_wifi_sta_pmksa_cache.c
sylvioalves: Yes

components/wpa_supplicant/esp_supplicant/src/esp_wifi_sta_pmksa_cache.c
```diff
@@ -0,0 +1,321 @@
...
+#include "common/defs.h"
+
+#include "esp_wifi_sta_pmksa_cache.h"
+#include "esp_wifi_sta_pmksa_cache_i.h"
+#include "rsn_supp/wpa.h"
+#include "rsn_supp/wpa_i.h"
+#include "rsn_supp/pmksa_cache.h"
+
+/* Record awaiting installation on the next association attempt. */
+static esp_wifi_sta_pmksa_cache_entry_t s_staged;
```
sylvioalves, line 18: one thing I missed to ask: I ma not sure how zephyr main repo PR will handle this but `s_staged/s_installed` are not reset on `esp_wifi_stop()` and  `wpa_sm_deinit()` frees the whole cache via `pmksa_cache_deinit()` without calling `free_cb`, so `s_installed.valid` can stay true for an entry that no longer exists, and the staged PMK stays in RAM. Should the module expose a reset that `wpa_sm_deinit()` calls?

siddhantmodi:
> would you also check whether we should set the expiration in? Otherwise flush or removal won't make expiry timer to re-arm again. https://github.com/zephyrproject-rtos/hal_espressif/blob/main/components/wpa_supplicant/src/rsn_supp/pmksa_cache.c#L306 https://github.com/zephyrproject-rtos/hal_espressif/blob/main/components/wpa_supplicant/src/rsn_supp/pmksa_cache.c#L80

I am not quite sure what you mean by this one. `pmksa_cache_set_expiration()` is static, not in `pmksa_cache.h`, so any changes will have to live in `pmksa_cache.c`. Are you suggesting that we uncomment [pmksa_cache.c/L306](https://github.com/zephyrproject-rtos/hal_espressif/blob/main/components/wpa_supplicant/src/rsn_supp/pmksa_cache.c#L306)? The [equivalent line](https://github.com/zephyrproject-rtos/hostap/blob/d467a115b8fc20b8b7120937cb1d445dacebc6a1/src/rsn_supp/pmksa_cache.c#L442) in upstream `hostap` is indeed commented out. I can add it as a separate commit in this PR.

sylvioalves:
> > would you also check whether we should set the expiration in? Otherwise flush or removal won't make expiry timer to re-arm again. https://github.com/zephyrproject-rtos/hal_espressif/blob/main/components/wpa_supplicant/src/rsn_supp/pmksa_cache.c#L306 https://github.com/zephyrproject-rtos/hal_espressif/blob/main/components/wpa_supplicant/src/rsn_supp/pmksa_cache.c#L80
> 
> I am not quite sure what you mean by this one. `pmksa_cache_set_expiration()` is static, not in `pmksa_cache.h`, so any changes will have to live in `pmksa_cache.c`. Are you suggesting that we uncomment [pmksa_cache.c/L306](https://github.com/zephyrproject-rtos/hal_espressif/blob/main/components/wpa_supplicant/src/rsn_supp/pmksa_cache.c#L306)? The [equivalent line](https://github.com/zephyrproject-rtos/hostap/blob/d467a115b8fc20b8b7120937cb1d445dacebc6a1/src/rsn_supp/pmksa_cache.c#L442) in upstream `hostap` is indeed commented out. I can add it as a separate commit in this PR.

Yes, we can/should uncomment it as done in the hostap upstream file.

sylvioalves:
> > > would you also check whether we should set the expiration in? Otherwise flush or removal won't make expiry timer to re-arm again. https://github.com/zephyrproject-rtos/hal_espressif/blob/main/components/wpa_supplicant/src/rsn_supp/pmksa_cache.c#L306 https://github.com/zephyrproject-rtos/hal_espressif/blob/main/components/wpa_supplicant/src/rsn_supp/pmksa_cache.c#L80
> > 
> > 
> > I am not quite sure what you mean by this one. `pmksa_cache_set_expiration()` is static, not in `pmksa_cache.h`, so any changes will have to live in `pmksa_cache.c`. Are you suggesting that we uncomment [pmksa_cache.c/L306](https://github.com/zephyrproject-rtos/hal_espressif/blob/main/components/wpa_supplicant/src/rsn_supp/pmksa_cache.c#L306)? The [equivalent line](https://github.com/zephyrproject-rtos/hostap/blob/d467a115b8fc20b8b7120937cb1d445dacebc6a1/src/rsn_supp/pmksa_cache.c#L442) in upstream `hostap` is indeed commented out. I can add it as a separate commit in this PR.
> 
> Yes, we can/should uncomment it as done in the hostap upstream file.

Unless you see something that would break in this feature bring up.. in this case we could do a follow up with proper/additional tests.

siddhantmodi:
@sylvioalves I think the change to `pmksa_cache.c` should be done as a separate patch. It is not related to this PR.

From looking at the code:

`dot11RSNAConfigPMKLifetime` is `INT32_MAX` here, so entries learned by this supplicant effectively never expire client-side, and this fork has no `pmksa_cache_reauth()`, so `entry->reauth_time` is written but never read.

I think this fix will involve restoring the reauth machinery, which is a large change.

sylvioalves:
@marekmatej @raffarost PTAL

sylvioalves (approved):

siddhantmodi:
@marekmatej @raffarost Please take a look when you have a chance, thank you!

sylvioalves:

sylvioalves:
Keeping DNM until https://github.com/zephyrproject-rtos/zephyr/pull/115470 gets approved.

siddhantmodi:
@sylvioalves Just FYI - I'm back from vacation now and will be working to resolve my open PRs. Thanks for keeping an eye on this while I was away.

tmedicci (approved):

PR #625, opened 2026-07-31
esp32c61: add initial support

sylvioalves:
This PR adds ESP32-C61 support to hal_espressif: the Zephyr port
layer, the Zephyr-build adaptations to the C61 components, and the
Wi-Fi and BLE integration with their prebuilt libraries.

### DIFF
--- a/components/esp_hal_clock/esp32c61/include/hal/clk_tree_ll.h
+++ b/components/esp_hal_clock/esp32c61/include/hal/clk_tree_ll.h
@@ -22,7 +22,13 @@
 #include "esp32c61/rom/rtc.h"
 #include "hal/misc.h"
 
-#define MHZ                 (1000000)
+/* Zephyr's sys/util.h provides a function-like MHZ(x) macro; use the
+ * same form here to avoid a collision with the original object-like
+ * MHZ definition.
+ */
+#ifndef MHZ
+#define MHZ(x)              ((x) * 1000000UL)
+#endif
 
 #define CLK_LL_PLL_80M_FREQ_MHZ    (80)
 #define CLK_LL_PLL_120M_FREQ_MHZ   (120)
--- a/components/esp_hal_dma/esp32c61/include/hal/gdma_ll.h
+++ b/components/esp_hal_dma/esp32c61/include/hal/gdma_ll.h
@@ -18,6 +18,46 @@
 #define GDMA_LL_AHB_BURST_SIZE_ADJUSTABLE 1  // AHB GDMA supports adjustable burst size
 #define GDMA_LL_MAX_BURST_SIZE_PSRAM      32 // PSRAM controller doesn't support burst access with size > 32 bytes
 
+/*
+ * Compatibility aliases: map gdma_ll_* names to the ahb_dma_ll_* API
+ * so that Zephyr drivers written against the v1 GDMA interface compile on
+ * AHB-GDMA-v2 SoCs without source changes.
+ */
+#define GDMA_LL_M2M_FREE_PERIPH_ID_MASK          AHB_DMA_LL_M2M_FREE_PERIPH_ID_MASK
+#define GDMA_LL_RX_EVENT_MASK                     AHB_DMA_LL_RX_EVENT_MASK
+#define GDMA_LL_TX_EVENT_MASK                     AHB_DMA_LL_TX_EVENT_MASK
+#define gdma_ll_force_enable_reg_clock            ahb_dma_ll_force_enable_reg_clock
+#define gdma_ll_rx_clear_interrupt_status          ahb_dma_ll_rx_clear_interrupt_status
+#define gdma_ll_rx_connect_to_mem                  ahb_dma_ll_rx_connect_to_mem
+#define gdma_ll_rx_connect_to_periph               ahb_dma_ll_rx_connect_to_periph
+#define gdma_ll_rx_enable_data_burst               ahb_dma_ll_rx_enable_data_burst
+#define gdma_ll_rx_enable_descriptor_burst         ahb_dma_ll_rx_enable_descriptor_burst
+#define gdma_ll_rx_enable_interrupt                ahb_dma_ll_rx_enable_interrupt
+#define gdma_ll_rx_enable_owner_check              ahb_dma_ll_rx_enable_owner_check
+#define gdma_ll_rx_get_interrupt_status            ahb_dma_ll_rx_get_interrupt_status
+#define gdma_ll_rx_get_interrupt_status_reg        ahb_dma_ll_rx_get_interrupt_status_reg
+#define gdma_ll_rx_get_prefetched_desc_addr        ahb_dma_ll_rx_get_prefetched_desc_addr
+#define gdma_ll_rx_get_success_eof_desc_addr       ahb_dma_ll_rx_get_success_eof_desc_addr
+#define gdma_ll_rx_is_desc_fsm_idle                ahb_dma_ll_rx_is_desc_fsm_idle
+#define gdma_ll_rx_reset_channel                   ahb_dma_ll_rx_reset_channel
+#define gdma_ll_rx_set_desc_addr                   ahb_dma_ll_rx_set_desc_addr
+#define gdma_ll_rx_start                           ahb_dma_ll_rx_start
+#define gdma_ll_rx_stop                            ahb_dma_ll_rx_stop
+#define gdma_ll_tx_clear_interrupt_status          ahb_dma_ll_tx_clear_interrupt_status
+#define gdma_ll_tx_connect_to_mem                  ahb_dma_ll_tx_connect_to_mem
+#define gdma_ll_tx_connect_to_periph               ahb_dma_ll_tx_connect_to_periph
+#define gdma_ll_tx_enable_data_burst               ahb_dma_ll_tx_enable_data_burst
+#define gdma_ll_tx_enable_descriptor_burst         ahb_dma_ll_tx_enable_descriptor_burst
+#define gdma_ll_tx_enable_interrupt                ahb_dma_ll_tx_enable_interrupt
+#define gdma_ll_tx_get_interrupt_status            ahb_dma_ll_tx_get_interrupt_status
+#define gdma_ll_tx_get_interrupt_status_reg        ahb_dma_ll_tx_get_interrupt_status_reg
+#define gdma_ll_tx_get_prefetched_desc_addr        ahb_dma_ll_tx_get_prefetched_desc_addr
+#define gdma_ll_tx_is_desc_fsm_idle                ahb_dma_ll_tx_is_desc_fsm_idle
+#define gdma_ll_tx_reset_channel                   ahb_dma_ll_tx_reset_channel
+#define gdma_ll_tx_set_desc_addr                   ahb_dma_ll_tx_set_desc_addr
+#define gdma_ll_tx_start                           ahb_dma_ll_tx_start
+#define gdma_ll_tx_stop                            ahb_dma_ll_tx_stop
+
 #ifdef __cplusplus
 extern "C" {
 #endif
--- a/components/esp_hal_i2s/esp32c61/include/hal/i2s_ll.h
+++ b/components/esp_hal_i2s/esp32c61/include/hal/i2s_ll.h
@@ -1372,6 +1372,7 @@ static inline bool i2s_ll_get_etm_rx_threshold_event_status(i2s_dev_t *hw)
     }
 }
 
+#ifndef __ZEPHYR__
 /**
  * @brief Set I2S data destination
  */
@@ -1390,6 +1391,7 @@ static inline bool i2s_ll_is_destination_supported(int port_id, i2s_destination_
     (void)port_id;
     return destination == I2S_DESTINATION_DMA;
 }
+#endif /* __ZEPHYR__ */
 
 /**
  * @brief Check whether I2S PDM mode is supported on the specified port
--- a/components/esp_hal_uart/esp32c61/include/hal/uart_ll.h
+++ b/components/esp_hal_uart/esp32c61/include/hal/uart_ll.h
@@ -1022,6 +1022,18 @@ FORCE_INLINE_ATTR void uart_ll_set_mode_rs485_half_duplex(uart_dev_t *hw)
 }
 
 /**
+ * @brief  Check if the UART is in rs485 half duplex mode.
+ *
+ * @param  hw Beginning address of the peripheral registers.
+ *
+ * @return True if the UART is in rs485 half duplex mode, otherwise false.
+ */
+FORCE_INLINE_ATTR bool uart_ll_is_mode_rs485_half_duplex(uart_dev_t *hw)
+{
+    return hw->rs485_conf_sync.rs485_en && hw->conf0_sync.sw_rts;
+}
+
+/**
  * @brief  Configure the UART work in collision_detect mode.
  *
  * @param  hw Beginning address of the peripheral registers.
--- a/components/esp_hw_support/port/esp32c61/esp_clk_tree.c
+++ b/components/esp_hw_support/port/esp32c61/esp_clk_tree.c
@@ -5,7 +5,7 @@
  */
 
 #include <stdint.h>
-#include <stdatomic.h>
+#include <zephyr/sys/atomic.h>
 #include "esp_clk_tree.h"
 #include "esp_err.h"
 #include "esp_check.h"
@@ -18,7 +18,7 @@
 
 ESP_LOG_ATTR_TAG(TAG, "esp_clk_tree");
 
-static _Atomic int16_t s_pll_src_cg_ref_cnt[SOC_MOD_CLK_INVALID] = { 0 };
+static atomic_t s_pll_src_cg_ref_cnt[SOC_MOD_CLK_INVALID];
 
 esp_err_t esp_clk_tree_src_get_freq_hz(soc_module_clk_t clk_src, esp_clk_tree_src_freq_precision_t precision,
 uint32_t *freq_value)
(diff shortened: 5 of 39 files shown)


components/esp_hal_dma/esp32c61/include/hal/gdma_ll.h
```diff
@@ -18,6 +18,46 @@
...
+#define gdma_ll_tx_enable_interrupt                ahb_dma_ll_tx_enable_interrupt
+#define gdma_ll_tx_get_interrupt_status            ahb_dma_ll_tx_get_interrupt_status
+#define gdma_ll_tx_get_interrupt_status_reg        ahb_dma_ll_tx_get_interrupt_status_reg
+#define gdma_ll_tx_get_prefetched_desc_addr        ahb_dma_ll_tx_get_prefetched_desc_addr
+#define gdma_ll_tx_is_desc_fsm_idle                ahb_dma_ll_tx_is_desc_fsm_idle
+#define gdma_ll_tx_reset_channel                   ahb_dma_ll_tx_reset_channel
+#define gdma_ll_tx_set_desc_addr                   ahb_dma_ll_tx_set_desc_addr
+#define gdma_ll_tx_start                           ahb_dma_ll_tx_start
+#define gdma_ll_tx_stop                            ahb_dma_ll_tx_stop
+
```
tmedicci, line 60: Why is this necessary? Zephyr already compiles [`gdma_hal_ahb_v2.c`](https://github.com/zephyrproject-rtos/hal_espressif/blob/aeef296e87e9d822259ac9a7402d52f757971ef5/zephyr/esp32c5/CMakeLists.txt#L315), which provides such a kind of abstraction layer.

components/esp_hal_i2s/esp32c61/include/hal/i2s_ll.h
```diff
@@ -1390,6 +1391,7 @@ static inline bool i2s_ll_is_destination_supported(int port_id, i2s_destination_
     (void)port_id;
     return destination == I2S_DESTINATION_DMA;
 }
+#endif /* __ZEPHYR__ */
```
tmedicci, line 1394: Why can't it be defined on Zephyhr?

components/esp_hal_uart/esp32c61/include/hal/uart_ll.h
```diff
@@ -1021,6 +1021,18 @@ FORCE_INLINE_ATTR void uart_ll_set_mode_rs485_half_duplex(uart_dev_t *hw)
...
+ * @brief  Check if the UART is in rs485 half duplex mode.
+ *
+ * @param  hw Beginning address of the peripheral registers.
+ *
+ * @return True if the UART is in rs485 half duplex mode, otherwise false.
+ */
+FORCE_INLINE_ATTR bool uart_ll_is_mode_rs485_half_duplex(uart_dev_t *hw)
+{
+    return hw->rs485_conf_sync.rs485_en && hw->conf0_sync.sw_rts;
+}
```
tmedicci, line 1034: Should we submit it to ESP-IDF so that you can reuse it on newer chips?

components/esp_hw_support/port/esp32c61/esp_clk_tree.c
```diff
@@ -99,12 +99,12 @@ esp_err_t esp_clk_tree_enable_src(soc_module_clk_t clk_src, bool enable)
     }
     int16_t prev_ref_cnt = 0;
     if (enable) {
-        prev_ref_cnt = atomic_fetch_add(&s_pll_src_cg_ref_cnt[clk_src], 1);
+        prev_ref_cnt = atomic_add(&s_pll_src_cg_ref_cnt[clk_src], 1);
```
tmedicci, line 105: Why is it necessary? Should we add a comment here explaining why the change is necessary (eventually submitting it to ESP-IDF)?

components/esp_hal_clock/esp32c61/include/hal/clk_tree_ll.h
```diff
@@ -22,7 +22,9 @@
 #include "esp32c61/rom/rtc.h"
 #include "hal/misc.h"
 
-#define MHZ                 (1000000)
+#ifndef MHZ
+#define MHZ(x)              ((x) * 1000000UL)
+#endif
```
tmedicci, line 31: What is the point of redefining it? Should we add a comment about why it is necessary?

components/esp_rom/esp32c61/ld/esp32c61.rom.ld
```diff
@@ -215,7 +215,7 @@ Cache_Suspend_Cache = 0x40000698;
 Cache_Resume_Cache = 0x4000069c;
 Cache_Freeze_Enable = 0x400006a0;
 Cache_Freeze_Disable = 0x400006a4;
-Cache_Set_IDROM_MMU_Size = 0x400006a8;
+PROVIDE ( Cache_Set_IDROM_MMU_Size = 0x400006a8 );
```
tmedicci: If it can't be used, better to remove it completely.

components/soc/esp32c61/include/soc/ext_mem_defs.h
```diff
@@ -133,6 +133,25 @@ extern "C" {
...
+#define CACHE_DROM_MMU_SIZE             (CACHE_DROM_MMU_END - CACHE_DROM_MMU_START)
+
+#define CACHE_DROM_MMU_MAX_END          0x400
+
+#define ICACHE_MMU_SIZE                 0x200
+#define DCACHE_MMU_SIZE                 0x200
+
+#define MMU_BUS_START(i)                0
+#define MMU_BUS_SIZE(i)                 0x200
+
```
tmedicci: Where are these definitions actually being used? I couldn't find anything on ESP-IDF (or current Zephyhr) that uses it...

zephyr/esp32c61/src/soc_random.c
tmedicci, line 1: It seems that this set of functions is identical to the ones at https://github.com/zephyrproject-rtos/hal_espressif/blob/46d603e1602d891e95d9c2bb70f23f8657aebf82/components/bootloader_support/src/bootloader_random_esp32c61.c

Why not using them directly?

zephyr/scripts/pinctrl/esp_genpinctrl.py
tmedicci, line 1: Wouldn't it apply to every chip? Is that intentional?

components/esp_wifi/esp32c61/esp_adapter.c
```diff
@@ -397,41 +566,187 @@ static int get_time_wrapper(void *t)
 
 static void *IRAM_ATTR realloc_internal_wrapper(void *ptr, size_t size)
 {
-    return heap_caps_realloc(ptr, size, MALLOC_CAP_8BIT | MALLOC_CAP_DMA | MALLOC_CAP_INTERNAL);
+    return heap_caps_realloc(ptr, size, MALLOC_CAP_8BIT | MALLOC_CAP_INTERNAL);
```
tmedicci, line 569: I just figured out that we have a `heap_caps` wrapper at https://github.com/zephyrproject-rtos/hal_espressif/blob/3a8e5fa3a93bb4c4b97d942a0a0ba3254733e5d6/zephyr/port/heap/heap_caps_zephyr.c

In this case, wouldn't it be better to either adopt or deprecate it fully? I mean: we have mixed usage in this file.

components/esp_hal_dma/esp32c61/include/hal/gdma_ll.h
sylvioalves: @tmedicci the Zephyr drivers (dma_esp32_gdma.c and uart_esp32.c) call the
gdma_ll_* names directly, not the gdma_hal_* layer provided by gdma_hal_ahb_v2.c. On AHB-GDMA-v2 SoCs the low level API was renamed to ahb_dma_ll_*, so without these aliases the drivers do not compile. This mirrors the same alias block already merged in components/esp_hal_dma/esp32c5/include/hal/gdma_ll.h.
Without those I would need to split dma drivers in Zephyr side, replicating too much code. Thoughts?

components/esp_hal_i2s/esp32c61/include/hal/i2s_ll.h
sylvioalves: humm, the guard is `#ifndef __ZEPHYR__` as these `i2s_ll_set_destination` stubs are excluded from Zephyr builds because the Zephyr I2S driver does not use the destination mux concept. They are kept for sync parity with the original header. The merged esp32c5 `i2s_ll.h` carries the identical block.

components/esp_hal_uart/esp32c61/include/hal/uart_ll.h
sylvioalves: yes, we should! For quite a long time now.

components/esp_hw_support/port/esp32c61/esp_clk_tree.c
sylvioalves: sure, I'll add a commnetn, it is worth. But the thing is: this is built using `<zephyr/sys/atomic.h>`
instead of C11 stdatomic. Zephyr's atomic_add returns the previous value, the same semantics as
atomic_fetch_add..

components/esp_hal_clock/esp32c61/include/hal/clk_tree_ll.h
sylvioalves: will add a comment in every `clk_tree_ll.h`. Yes, it conflicts with Zephyr's MHz definition.

zephyr/scripts/pinctrl/esp_genpinctrl.py
sylvioalves: yes, will remove, thanks.

components/esp_wifi/esp32c61/esp_adapter.c
sylvioalves: This file is identical to the merged esp32c5 and esp32c6 adapters in this respect, so fixing it only for esp32c61 would break cross-SoC parity and complicate future syncs. I would rather do a dedicated cleanup in all esp_adapter.c files as a follow-up PR, do you mind?

components/esp_hal_dma/esp32c61/include/hal/gdma_ll.h
tmedicci: Alright, for [`uart_esp32.c`](https://github.com/zephyrproject-rtos/zephyr/blob/f9506e67ac5a092f9f8968f12a43d54d013fbc59/drivers/serial/uart_esp32.c#L737)'s `gdma_ll_rx_get_success_eof_desc_addr` we could use HAL's [`gdma_ahb_hal_get_eof_desc_addr`](https://github.com/zephyrproject-rtos/hal_espressif/blob/46d603e1602d891e95d9c2bb70f23f8657aebf82/components/esp_hal_dma/gdma_hal_ahb_v1.c#L163) directly. It's the safest option.

About `dma_esp32_gdma.c`, I've seen that other SoCs do the same. For now, by comparing `components/esp_hal_dma/gdma_hal_ahb_v1.c` and `components/esp_hal_dma/gdma_hal_ahb_v2.c` side-by-side, the `_ll` APIs seem to correspond 1:1, but we should take it as the ground truth. So, for now, I think we are fine keeping it the way it's, but we should slowly migrate to using the `esp_hal_` interfaces to avoid any silent bugs introduced by this 1:1 correspondence.

components/esp_hal_i2s/esp32c61/include/hal/i2s_ll.h
tmedicci: I see, but my point is more about: why not keep it here and not introduce any unnecessary changes to the HAL code? I meant: if this isn't being used anywhere, it should do no harm, right?

components/esp_hw_support/port/esp32c61/esp_clk_tree.c
```diff
@@ -5,7 +5,7 @@
  */
 
 #include <stdint.h>
-#include <stdatomic.h>
+#include <zephyr/sys/atomic.h>
```
tmedicci, line 8: Why can't we use stdatomic here? (not a blocker, just for the sake of curiosity)

components/esp_hw_support/port/esp32c61/esp_clk_tree.c
tmedicci: ow, already asked on another comment, but what are the advantages of using Zephyr's atomic add? I meant: if we couldn't avoid any unnecessary changes on HAL, it'd be better. So, I'm just curious why Zephyr's atomic is desirable here.

tmedicci (approved):
None of my comments are blockers to it (they are closer to suggestions/information requests).

components/esp_rom/esp32c61/ld/esp32c61.rom.ld
sylvioalves: During bring up, I spot issues when using this and left as PROVIDE without any need. Will remove.

zephyr/esp32c61/src/soc_random.c
sylvioalves: hum, almost identical.. `soc_random.c` is that file with the `BOOTLOADER_BUILD` branches resolved for the Zephyr runtime. It uses the ref-counted `regi2c_saradc_enable/disable` because the locking wrappers inside `sar_periph_ctrl_adc_reset()` are not available at the early loader stage where `soc_random_enable()` runs. So instead of changing the `bootloader_random-xxx` file, each soc carries its own `soc_random.c` zephyr-compatible style. Now, we can improve this eventually.

zephyr/scripts/pinctrl/esp_genpinctrl.py
sylvioalves: This snippet of `uart` entry gets pulled in all SoC as is, which is wrong and we can't filter out per soc. So I'll remove the "example" from here, intentionally.

components/esp_hw_support/port/esp32c61/esp_clk_tree.c
sylvioalves: simple: Zephyr does not build or link libatomic, so any C11 atomic op GCC cannot inline becomes an undefined __atomic_* reference.

components/soc/esp32c61/include/soc/ext_mem_defs.h
sylvioalves: removed, thanks.

components/esp_hw_support/port/esp32c61/esp_clk_tree.c
sylvioalves: replied in the other comment

components/esp_hal_i2s/esp32c61/include/hal/i2s_ll.h
sylvioalves: `i2s_dir_t` is both defined in Zephyr (i2s.h) and in hal_espressif's (i2s_ll.h). The only function that uses `i2s_dir` is `i2s_ll_set_destination`, so we have to pull it out of hal, otherwise build fails.

sylvioalves: Do you mind if I keep the #ifdef as done in all socs already?

components/esp_wifi/esp32c61/esp_adapter.c
tmedicci: Nope, totally reasonable.

zephyr/esp32c61/include/sdkconfig.h
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright (c) 2026 Espressif Systems (Shanghai) Co., Ltd.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef CONFIG_BOOTLOADER_OFFSET_IN_FLASH
+#define CONFIG_BOOTLOADER_OFFSET_IN_FLASH 0x0
```
raffarost: maybe:

```c
config BOOTLOADER_OFFSET_IN_FLASH
	hex
	default $(dt_node_reg_addr_hex,$(dt_nodelabel_path,boot_partition))
```

is safer? also for C5

zephyr/port/pincfgs/esp32c61.yml
```diff
@@ -0,0 +1,154 @@
+# Copyright (c) 2026 Espressif Systems (Shanghai) Co., Ltd.
+# SPDX-License-Identifier: Apache-2.0
+
+# Notes:
+# - ESP32-C61 has 30 GPIO pads (GPIO0-GPIO29)
+# - GPIO15-GPIO21 are the MSPI flash pads (CS0, MISO, WP, VDD_SPI, HD,
+#   CLK, MOSI) and are excluded from the peripheral mux ranges
```
raffarost, line 7: Reminder to add this to the `.dtsi` (gpio0 node):

`gpio-reserved-ranges = <15 7>; /* GPIO15–21 MSPI flash (CS0..MOSI) */`

raffarost (approved):
I couldn't help but notice how many duplicated functions we have in HAL (e.g. esp_adapter). Would be interesting to use `__weak` directives and common files to only specialize code when needed, but this would be a HAL/IDF improvement, so not related to this PR.

zephyr/common/console_init.c
```diff
@@ -10,7 +10,9 @@
 #include "soc/io_mux_reg.h"
 #include "soc/gpio_periph.h"
 #include "soc/gpio_sig_map.h"
+#if !defined(CONFIG_SOC_SERIES_ESP32C61)
 #include "hal/clk_gate_ll.h"
```
marekmatej: Instead of gating, could we mock up this file as in case of `esp_key_mgr.h`

zephyr/esp32c61/CMakeLists.txt
```diff
@@ -0,0 +1,801 @@
...
+
+  zephyr_include_directories(
+    include
+    ../common/include
+    ../port/include
+    ../port/include/boot
+    ../../components/bootloader_support/bootloader_flash/include
+    ../../components/bootloader_support/include
+    ../../components/bootloader_support/private_include
+    ## esp32c61 reuses the esp32c6 ble controller sources and headers
```
marekmatej: single # in comment

zephyr/scripts/pinctrl/esp_genpinctrl.py
```diff
@@ -48,22 +48,6 @@
  * For example, `I2C0_SCL_GPIO1` corresponds to the `SCL` signal of `I2C0`
  * mapped to `GPIO1`.
  *
- * @code{.dts}
```
marekmatej, line 51: why removing the example? I'd update it instead, up to you

marekmatej (commented):
nits, ott lgtm

zephyr/esp32c61/include/sdkconfig.h
sylvioalves: yea, this was leftover and should go.

zephyr/port/pincfgs/esp32c61.yml
sylvioalves: nice, will take it.

zephyr/esp32c61/include/sdkconfig.h
sylvioalves: addressed

sylvioalves:
> I couldn't help but notice how many duplicated functions we have in HAL (e.g. esp_adapter). Would be interesting to use `__weak` directives and common files to only specialize code when needed, but this would be a HAL/IDF improvement, so not related to this PR.

Sure

zephyr/esp32c61/CMakeLists.txt
sylvioalves: addressed

zephyr/scripts/pinctrl/esp_genpinctrl.py
sylvioalves: well, any fixed pin number in a template shared by all chips is a trap. See, GPIO21/20 already do not exist on
esp32c61 so would be wrong to have it here. And most pinctrl dt-bindings headers in the tree carry no example.

zephyr/common/console_init.c
sylvioalves: Ends up that clk_ll and periph_ll are not used at ll in this file, so removed it.

marekmatej (approved):

PR #627, opened 2026-07-31
esp32c5: follow vendor analog and clock init

sylvioalves:
The rc fast clock tuning field was set to 172, a value carried over from an older soc; the esp32c5 default is 100, which is what the vendor clock init writes.

The mspi function clock source and its 80 MHz core clock were only configured for 80 MHz flash builds, so slower flash modes divided the wrong base clock. Set them for every flash mode once the pll is up, as the vendor does, and keep only the divider fix under the 80 MHz option.

Also arm the power glitch reset, which the vendor analog reset config enables next to the brownout reset.

### DIFF
--- a/zephyr/esp32c5/src/soc_init.c
+++ b/zephyr/esp32c5/src/soc_init.c
@@ -9,6 +9,7 @@
 #include "esp32c5/rom/rtc.h"
 #include <soc/soc.h>
 #include "soc/lp_analog_peri_reg.h"
+#include "soc/regi2c_saradc.h"
 #include "hal/clk_tree_ll.h"
 #include "hal/brownout_ll.h"
 #include "hal/regi2c_ctrl_ll.h"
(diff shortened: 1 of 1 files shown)


zephyr/esp32c5/src/soc_init.c
```diff
@@ -170,7 +182,7 @@ void bootloader_clock_configure(void)
 	/* Set tuning parameters for RC_FAST and RC_SLOW clocks
 	 * (matches ESP-IDF rtc_clk_init).
 	 */
-	REG_SET_FIELD(LP_CLKRST_FOSC_CNTL_REG, LP_CLKRST_FOSC_DFREQ, 172);
+	REG_SET_FIELD(LP_CLKRST_FOSC_CNTL_REG, LP_CLKRST_FOSC_DFREQ, 100);
```
raffarost: maybe use `RTC_CNTL_CK8M_DFREQ_DEFAULT` here too.

raffarost (approved):

wmrsouza (approved):

marekmatej (approved):

PR #632, opened 2026-08-12
zephyr: use rom printf for hal logs until kernel is up

sylvioalves:
Hal code can log from the boot loader stage, before the flash cache is mapped. With CONFIG_LOG enabled, ESP_LOGx expands to printk, which is redirected into the deferred logging path whose buffer code resides in flash-mapped memory.

Route ESP_LOGx through the ROM printf while k_is_pre_kernel() is true and through printk afterwards. The boot entry zeroes bss before the loader runs, so the check is reliable in every boot stage and covers all current and future hal log call sites.

### DIFF
--- a/zephyr/common/include/esp_log.h
+++ b/zephyr/common/include/esp_log.h
@@ -9,6 +9,7 @@
 #include <stdarg.h>
 #include <stdint.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


raffarost (approved):

marekmatej (approved):
nice!

PR #642, opened 2026-09-02
zephyr: build the aes and sha hal per enabled engine

sylvioalves:
CONFIG_CRYPTO_ESP32 pulled in both aes_hal.c and sha_hal.c, which breaks socs such as the esp32c61 that only have a SHA engine and therefore ship no aes_ll.h. Compile each hal behind the driver symbol of its own engine instead.


sylvioalves:

raffarost (approved):

tmedicci (approved):

PR #644, opened 2026-09-02
esp_hw_support: prepare the shared hal paths for smp

sylvioalves:
Makes the hal paths both cores can reach SMP-safe (arch_irq_lock in pre-kernel code, cross-core park around flash cache suspend) and pins the esp_timer task to one core. Groundwork for the ESP32-P4 and ESP32-S3 SMP support in Zephyr; no change for single-core builds.


sylvioalves:

components/esp_timer/src/esp_timer.c
```diff
@@ -613,11 +613,24 @@ static esp_err_t init_timer_task(void)
...
                         timer_task, NULL, NULL, NULL,
-                        K_PRIO_PREEMPT(CONFIG_ESP32_TIMER_TASK_PRIO), 0, K_NO_WAIT);
+                        K_PRIO_PREEMPT(CONFIG_ESP32_TIMER_TASK_PRIO), 0,
+                        IS_ENABLED(CONFIG_SMP) ? K_FOREVER : K_NO_WAIT);
         k_thread_name_set(&s_timer_task, "esp_timer");
+
+#if defined(CONFIG_SMP)
+#ifdef CONFIG_SCHED_CPU_MASK
+        k_thread_cpu_pin(tid, CONFIG_SOC_ESP32_TIMER_TASK_CORE_ID);
+#endif
```
tmedicci, line 629: Woulnd't `CONFIG_SCHED_CPU_MASK` be set whenever `CONFIG_SMP` is true? If so, why would this snippet be optional?

Are there any case where SMP is enabled and `CONFIG_SCHED_CPU_MASK` isn't set (and, then, timer isn't being pinned to a specific core)?

components/spi_flash/cache_utils.c
```diff
@@ -37,6 +37,24 @@
...
+ */
+void soc_mp_pause_others(void);
+void soc_mp_resume_others(void);
+#define SPI_FLASH_PAUSE_OTHERS()    soc_mp_pause_others()
+#define SPI_FLASH_RESUME_OTHERS()   soc_mp_resume_others()
+#else
+#define SPI_FLASH_PAUSE_OTHERS()
+#define SPI_FLASH_RESUME_OTHERS()
+#endif
+
```
tmedicci, line 57: This doesn't seem to be true: on SMP, the other core always needs to be paused regardless of the cache being per-core (as on ESP32) or shared. The main point here is that this is being used by SPI flash operations and we need to ensure that the other core isn't executing code from the flash.

The conclusion here is that `soc_mp_pause_others` is always necessary. Can you please double-check it?

components/esp_timer/src/esp_timer.c
sylvioalves: CONFIG_SCHED_CPU_MASK is not implied by CONFIG_SMP. It is a standalone option in kernel/smp/Kconfig with no select or default y coming from SMP, and its help text calls out the extra scheduler overhead, so it stays off unless something actually asks for CPU affinity. A plain P4 build with CONFIG_SMP=y and no affinity user has it unset, and k_thread_cpu_pin() would not compile in that configuration. When it is off the timer task just runs unpinned, which is the only thing possible without affinity support in the kernel.

The outer CONFIG_SMP guard is doing something different, which is easy to miss: under SMP the task is created suspended so it can be pinned before it runs, then started explicitly. Without SMP it is created running. That asymmetry is why there are two guards rather than one.

components/spi_flash/cache_utils.c
sylvioalves: You are right, the comment justified this on the wrong grounds. The reason to park the peer is not that the cache is shared, it is that a flash operation suspends the cache and no core may be executing flash-mapped code while that is happening. That holds regardless of cache topology. I have reworded both the comment and the commit body to say that.

I kept the guard itself as is. soc_mp_pause_others() comes from each SoC's SMP support, and P4 and S3 are the two that implement it today: P4 builds esp_mp_stall.c on the SMP branch, and so does the S3 SMP branch. Widening the condition to a SoC that has no SMP port yet would just be an undefined reference. The new comment says that explicitly, so whoever adds the next SoC knows the guard has to grow with it.

components/esp_timer/src/esp_timer.c
tmedicci: Alright, thanks!

components/spi_flash/cache_utils.c
tmedicci: In this case, @sylvioalves , wouldn't the `CONFIG_SMP` check enough? I meant: SMP isn't implemented on ESP32, right? That being said, there is no need to 

```
#if defined(CONFIG_SMP) && \
	(defined(CONFIG_SOC_SERIES_ESP32P4) || defined(CONFIG_SOC_SERIES_ESP32S3))
```

On the other hand, 
```
#ifdef CONFIG_SMP
```

would be enough, isn't it?

tmedicci (approved):
Please check about `#ifdef CONFIG_SMP` comment (it's a redundancy issue, only), not a blocker. I'll let it approved.

sylvioalves:
@tmedicci applied. @raffarost PTAL.

marekmatej (approved):

PR #645, opened 2026-09-02
esp32s3: pin the radio tasks and reserve their interrupt lines for smp

sylvioalves:
The Wi-Fi and Bluetooth blobs expect their task and interrupt on the same core, so pin those tasks under SMP and reserve the Wi-Fi interrupt lines on both cores instead of one. Depends on the SMP groundwork PR; verified on ESP32-S3 with the Wi-Fi task on either core.

### DIFF
--- a/components/bt/controller/esp32/bt.c
+++ b/components/bt/controller/esp32/bt.c
@@ -833,11 +833,23 @@ static int32_t IRAM_ATTR queue_recv_from_isr_wrapper(void *queue, void *item, vo
 
 static int32_t task_create_wrapper(void *task_func, const char *name, uint32_t stack_depth, void *param, uint32_t prio, void *task_handle, uint32_t core_id)
 {
+    ARG_UNUSED(core_id);
+
+    /* The controller asserts its task runs on the configured core, so pin it
+     * (create suspended, pin, then start) instead of letting the scheduler
+     * float it.
+     */
     k_tid_t tid = k_thread_create(&bt_task_handle, bt_stack, stack_depth,
                       (k_thread_entry_t)task_func, param, NULL, NULL,
-                      K_PRIO_COOP(prio), K_INHERIT_PERMS, K_NO_WAIT);
+                      K_PRIO_COOP(prio), K_INHERIT_PERMS,
+                      IS_ENABLED(CONFIG_SMP) ? K_FOREVER : K_NO_WAIT);
 
     k_thread_name_set(tid, name);
+
+#if defined(CONFIG_SMP)
+    k_thread_cpu_pin(tid, CONFIG_ESP32_BT_CTLR_PINNED_TO_CORE);
+    k_thread_start(tid);
+#endif
 
     *(int32_t *)task_handle = (int32_t)tid;
 
--- a/components/esp_hw_support/port/esp32s3/esp_cpu_intr.c
+++ b/components/esp_hw_support/port/esp32s3/esp_cpu_intr.c
@@ -22,16 +22,13 @@ typedef struct {
 
 const static intr_desc_t intr_desc_table [SOC_CPU_INTR_NUM] = {
     /* Interrupt 0 reserved for WMAC (Wifi) */
-#if CONFIG_ESP_WIFI_TASK_PINNED_TO_CORE_0
-    [0] = { 1, ESP_CPU_INTR_TYPE_LEVEL,  { ESP_CPU_INTR_DESC_FLAG_RESVD,   0                               } },
-#else
-    [0] = { 1, ESP_CPU_INTR_TYPE_LEVEL,  { 0,                              ESP_CPU_INTR_DESC_FLAG_RESVD    } },
-#endif
-    [1] = { 1, ESP_CPU_INTR_TYPE_LEVEL,  { 0,                              0                               } },
+    [0] = { 1, ESP_CPU_INTR_TYPE_LEVEL,  { ESP_CPU_INTR_DESC_FLAG_RESVD,   ESP_CPU_INTR_DESC_FLAG_RESVD    } },
+    /* Interrupt 1 used by the WiFi blob alongside int 0 */
+    [1] = { 1, ESP_CPU_INTR_TYPE_LEVEL,  { ESP_CPU_INTR_DESC_FLAG_RESVD,   ESP_CPU_INTR_DESC_FLAG_RESVD    } },
     [2] = { 1, ESP_CPU_INTR_TYPE_LEVEL,  { 0,                              0                               } },
     [3] = { 1, ESP_CPU_INTR_TYPE_LEVEL,  { 0,                              0                               } },
     /* Interrupt 4 reserved for WBB */
-    [4] = { 1, ESP_CPU_INTR_TYPE_LEVEL,  { ESP_CPU_INTR_DESC_FLAG_RESVD,   0                               } },
+    [4] = { 1, ESP_CPU_INTR_TYPE_LEVEL,  { ESP_CPU_INTR_DESC_FLAG_RESVD,   ESP_CPU_INTR_DESC_FLAG_RESVD    } },
     [5] = { 1, ESP_CPU_INTR_TYPE_LEVEL,  { 0,                              0                               } },
     [6] = { 1, ESP_CPU_INTR_TYPE_NA,     { ESP_CPU_INTR_DESC_FLAG_SPECIAL, ESP_CPU_INTR_DESC_FLAG_SPECIAL  } },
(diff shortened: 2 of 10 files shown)


components/esp_hw_support/port/esp32s3/esp_cpu_intr.c
tmedicci, line 1: I really didn't get why this commit is necessary. We need to pin the Wi-Fi task to a specific core (that's what `CONFIG_ESP_WIFI_TASK_PINNED_TO_CORE_0` does). On the other core, the same CPU interrupt line can be used by any other peripheral. It doesn't need to be reserved...

Apart from that, it seems "dangerous"  to change this file: it's usually the ground truth for CPU interrupt allocation on each SoC: editing it in `hal_espressif` may lead to erroneous conclusions.

components/bt/controller/esp32/bt.c
```diff
@@ -833,12 +833,24 @@ static int32_t IRAM_ATTR queue_recv_from_isr_wrapper(void *queue, void *item, vo
...
-                      K_PRIO_COOP(prio), K_INHERIT_PERMS, K_NO_WAIT);
+                      K_PRIO_COOP(prio), K_INHERIT_PERMS,
+                      IS_ENABLED(CONFIG_SMP) ? K_FOREVER : K_NO_WAIT);
 
     k_thread_name_set(tid, name);
 
+#if defined(CONFIG_SMP)
+    k_thread_cpu_pin(tid, CONFIG_ESP32_BT_CTLR_PINNED_TO_CORE);
+    k_thread_start(tid);
+#endif
```
tmedicci, line 852: It seems that `core_id` is defined by the same file's `btdm_config_mask_load` function, which already takes `CONFIG_ESP32_BT_CTLR_PINNED_TO_CORE` into consideration.

```suggestion
#if defined(CONFIG_SMP)
    k_thread_cpu_pin(tid, core_id);
    k_thread_start(tid);
#endif
```

components/bt/controller/esp32/bt.c
```diff
@@ -833,12 +833,24 @@ static int32_t IRAM_ATTR queue_recv_from_isr_wrapper(void *queue, void *item, vo
 
 static int32_t task_create_wrapper(void *task_func, const char *name, uint32_t stack_depth, void *param, uint32_t prio, void *task_handle, uint32_t core_id)
 {
+    ARG_UNUSED(core_id);
```
tmedicci, line 836: Remove it (explained later)

components/esp_timer/src/esp_timer.c
```diff
@@ -613,11 +613,24 @@ static esp_err_t init_timer_task(void)
...
                         timer_task, NULL, NULL, NULL,
-                        K_PRIO_PREEMPT(CONFIG_ESP32_TIMER_TASK_PRIO), 0, K_NO_WAIT);
+                        K_PRIO_PREEMPT(CONFIG_ESP32_TIMER_TASK_PRIO), 0,
+                        IS_ENABLED(CONFIG_SMP) ? K_FOREVER : K_NO_WAIT);
         k_thread_name_set(&s_timer_task, "esp_timer");
+
+#if defined(CONFIG_SMP)
+#ifdef CONFIG_SCHED_CPU_MASK
+        k_thread_cpu_pin(tid, CONFIG_SOC_ESP32_TIMER_TASK_CORE_ID);
+#endif
```
tmedicci, line 629: https://github.com/zephyrproject-rtos/hal_espressif/pull/644#discussion_r3980889593

components/esp_wifi/esp32s3/esp_adapter.c
tmedicci, line 441: Remove it (check next comment)

components/esp_wifi/esp32s3/esp_adapter.c
```diff
@@ -463,12 +463,22 @@ static int32_t task_create_pinned_to_core_wrapper(void *task_func, const char *n
...
-				      prio, K_INHERIT_PERMS, K_NO_WAIT);
+				      prio, K_INHERIT_PERMS,
+				      IS_ENABLED(CONFIG_SMP) ? K_FOREVER : K_NO_WAIT);
 
 	k_thread_name_set(tid, name);
 
+#if defined(CONFIG_SMP)
+	k_thread_cpu_pin(tid, IS_ENABLED(CONFIG_ESP_WIFI_TASK_PINNED_TO_CORE_1) ? 1 : 0);
+	k_thread_start(tid);
+#endif
```
tmedicci, line 480: Same: [`esp32_wifi_dev_init`](https://github.com/zephyrproject-rtos/zephyr/blob/69192d66b3910db6ae1080aa9139e80f4f6529d6/drivers/wifi/esp32/src/esp_wifi_drv.c#L1921) uses `WIFI_INIT_CONFIG_DEFAULT`, which defines `WIFI_TASK_CORE_ID`, whose definition is given by `CONFIG_ESP_WIFI_TASK_PINNED_TO_CORE_1`.

```suggestion
#if defined(CONFIG_SMP)
	k_thread_cpu_pin(tid, IS_ENABLED(core_id);
	k_thread_start(tid);
#endif
```

components/spi_flash/cache_utils.c
tmedicci, line 1: https://github.com/zephyrproject-rtos/hal_espressif/pull/644#discussion_r3981035817

components/esp_hw_support/port/esp32s3/esp_cpu_intr.c
sylvioalves: addressed

components/bt/controller/esp32/bt.c
sylvioalves: done

components/bt/controller/esp32/bt.c
sylvioalves: removed

components/esp_timer/src/esp_timer.c
sylvioalves: updated so this PR goes one step further than the answer on #644. The radio stacks expect their timers on the same core as their own tasks, so leaving the timer task unpinned when CONFIG_SCHED_CPU_MASK is off is not a valid SMP configuration here.

components/esp_wifi/esp32s3/esp_adapter.c
sylvioalves: removed

components/esp_wifi/esp32s3/esp_adapter.c
sylvioalves: done

components/spi_flash/cache_utils.c
sylvioalves: This PR no longer touches cache_utils.c.

components/esp_hw_support/port/esp32s3/esp_cpu_intr.c
```diff
@@ -22,7 +22,7 @@ typedef struct {
 
 const static intr_desc_t intr_desc_table [SOC_CPU_INTR_NUM] = {
     /* Interrupt 0 reserved for WMAC (Wifi) */
-#if CONFIG_ESP_WIFI_TASK_PINNED_TO_CORE_0
+#if CONFIG_ESP32_WIFI_TASK_PINNED_TO_CORE_0
```
tmedicci: Can't we just keep the original code and set it on based on `CONFIG_ESP32_WIFI_TASK_PINNED_TO_CORE_0` at `zephyr/port/include/zephyr_compat.h`. This way we avoid introducing unnecessary changes to `components` folder.

components/esp_wifi/include/esp_wifi.h
```diff
@@ -196,7 +196,7 @@ extern wifi_osi_funcs_t g_wifi_osi_funcs;
 #define WIFI_DEFAULT_RX_BA_WIN 0 /* unused if ampdu_rx_enable == false */
 #endif
 
-#if CONFIG_ESP_WIFI_TASK_PINNED_TO_CORE_1
+#if CONFIG_ESP32_WIFI_TASK_PINNED_TO_CORE_1
```
tmedicci: Same.

zephyr/Kconfig
```diff
@@ -137,6 +137,13 @@ config ESP_SYSTEM_SINGLE_CORE_MODE
 	bool
 	default y if !SMP
 
+config ESP_IPC_ISR_STALL
+	bool
+	default y if SMP && !SOC_SERIES_ESP32
```
tmedicci: I'm sorry If I'm being too repetitive, but why does it need `!SOC_SERIES_ESP32` here? As far as I understand, that would be needed by ESP32 as well. If SMP isn't yet supported on ESP32, we can just doesn't let it to be enabled there (and that'd be enough to prevent users from accidentally enabling this)

PR #647, opened 2026-09-03
zephyr: import blobs as targets instead of -l/-L

kartben:
Blobs are linked by bare library name plus a `-L` at the SoC's blob directory. A bare name is not a file as far as CMake is concerned, so nothing in the build graph depends on the archives: after `west blobs fetch` pulls updated blobs, or a module revision shipping different ones is checked out, the previously linked copies stay in the image because no relink is triggered. `-L` search also resolves very generic names — `core`, `pp`, `phy`, `rtc` — against whatever else is on the library search path.

Adds `zephyr/cmake/esp_blobs.cmake` with an `esp_blobs_link()` helper that imports each blob as a `STATIC IMPORTED` target and links it, so the link step depends on the archive. Same shape `hal_silabs` uses (`add_prebuilt_library`) and `hal_afbr`.

The imported targets are attached to `zephyr_interface` rather than `ZEPHYR_LIBS`, which is why `zephyr_library_import()` is not used: Zephyr links `ZEPHYR_LIBS` with `--whole-archive`, which would pull every object of `libnet80211.a` / `libpp.a` / `libphy.a` into the image.

Preserved: link order including `gcc` after the esp libs, all `-T` linker script arguments, and `CONFIG_BUILD_ONLY_NO_BLOBS` behaviour (the helper is a no-op there, stubs untouched).

## Testing

Every SoC's blob list was diffed against the pre-conversion source: identical set and identical order in all nine, and every name resolves to a `lib/<soc>/lib<name>.a` declared in `zephyr/module.yml`.

`samples/net/wifi/shell` on `esp32c5_devkitc//hpcore`, built against `main` and against this branch:

- Identical memory report, and `nm -S --defined-only` gives the same 8177 symbols at the same sizes — no `--whole-archive` or link-order regression.
- The link line drops `-lcore -lespnow -lmesh -lnet80211 -lphy -lpp` and the `-L .../blobs/lib/esp32c5`, and gains the six archives by absolute path. `-lgcc` is untouched.
- `touch`ing `libpp.a` and re-running ninja: `main` reports **no work to do** and keeps the stale blob in the image; this branch relinks and regenerates the image.

### DIFF
--- a/zephyr/cmake/esp_blobs.cmake
+++ b/zephyr/cmake/esp_blobs.cmake
@@ -1,0 +1,25 @@
+# SPDX-FileCopyrightText: Copyright The Zephyr Project Contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# esp_blobs_link(<name> [<name> ...])
+#
+# Import lib<name>.a from the blob directory of the SoC series being built and
+# link it, in the order given. No-op when building without blobs.
+function(esp_blobs_link)
+  if(CONFIG_BUILD_ONLY_NO_BLOBS)
+    return()
+  endif()
+
+  set(blobs_dir ${ZEPHYR_HAL_ESPRESSIF_MODULE_DIR}/zephyr/blobs/lib/${CONFIG_SOC_SERIES})
+
+  foreach(name ${ARGV})
+    if(NOT TARGET esp_blob_${name})
+      add_library(esp_blob_${name} STATIC IMPORTED GLOBAL)
+      set_target_properties(esp_blob_${name} PROPERTIES
+        IMPORTED_LOCATION ${blobs_dir}/lib${name}.a
+      )
+    endif()
+    zephyr_link_libraries(esp_blob_${name})
+  endforeach()
+endfunction()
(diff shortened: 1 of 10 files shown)


sylvioalves:
@kartben thanks!

zephyr/cmake/esp_blobs.cmake
```diff
@@ -0,0 +1,25 @@
...
+# Import lib<name>.a from the blob directory of the SoC series being built and
+# link it, in the order given. No-op when building without blobs.
+function(esp_blobs_link)
+  if(CONFIG_BUILD_ONLY_NO_BLOBS)
+    return()
+  endif()
+
+  set(blobs_dir ${ZEPHYR_HAL_ESPRESSIF_MODULE_DIR}/zephyr/blobs/lib/${CONFIG_SOC_SERIES})
+
+  foreach(name ${ARGV})
```
sylvioalves, line 16: perhaps ARGN fits better here - not blocking
```suggestion
  foreach(name ${ARGN})
```

sylvioalves (approved):

marekmatej (approved):

PR #650, opened 2026-09-15
pm: ieee802.15.4: Add PM support

raffarost:
Adapt HAL files to Zephyr standard to allow enabling sleep retention for IEEE802.15.4 PM support.
Replace mutexes taken from ISR context or with interrupts locked, with binary semaphores.


components/esp_phy/src/btbb_init.c
```diff
@@ -18,7 +18,7 @@ K_MUTEX_DEFINE(s_btbb_access_lock);
 static uint8_t s_btbb_access_ref = 0;
 
 
-#if SOC_PM_MODEM_RETENTION_BY_REGDMA && CONFIG_FREERTOS_USE_TICKLESS_IDLE
+#if SOC_PM_MODEM_RETENTION_BY_REGDMA && CONFIG_PM
```
sylvioalves: bare` CONFIG_PM` also turns this on for ble-only builds, doesnt it? could it match the `sleep_clock.c` guard `(MAC_BB_PD || BT_LE_SLEEP_ENABLE || IEEE802154_SLEEP_ENABLE)` so btbb only registers when CLOCK_MODEM does? (same for all #ifdefs)

```suggestion
#if SOC_PM_MODEM_RETENTION_BY_REGDMA && (CONFIG_MAC_BB_PD || CONFIG_BT_LE_SLEEP_ENABLE || CONFIG_IEEE802154_SLEEP_ENABLE)
```

raffarost: Thanks, good catch

components/esp_phy/src/btbb_init.c
```diff
@@ -18,7 +18,7 @@ K_MUTEX_DEFINE(s_btbb_access_lock);
 static uint8_t s_btbb_access_ref = 0;
 
 
-#if SOC_PM_MODEM_RETENTION_BY_REGDMA && CONFIG_FREERTOS_USE_TICKLESS_IDLE
+#if SOC_PM_MODEM_RETENTION_BY_REGDMA && (CONFIG_MAC_BB_PD || CONFIG_BT_LE_SLEEP_ENABLE || CONFIG_IEEE802154_SLEEP_ENABLE)
```
marekmatej: perhaps you missed the `CONFIG_BT_CTRL_SLEEP_ENABLE`?

components/ieee802154/driver/esp_ieee802154_dev.c
```diff
@@ -1088,11 +1086,11 @@ esp_err_t ieee802154_receive_at(uint32_t time, uint32_t duration)
     return ESP_OK;
 }
 
-#if CONFIG_PM_ENABLE
+#if CONFIG_IEEE802154_SLEEP_ENABLE
 static esp_err_t ieee802154_sleep_retention_init(void *arg)
 {
     esp_err_t err = ESP_OK;
-#if SOC_PM_MODEM_RETENTION_BY_REGDMA && CONFIG_FREERTOS_USE_TICKLESS_IDLE
+#if SOC_PM_MODEM_RETENTION_BY_REGDMA && CONFIG_IEEE802154_SLEEP_ENABLE
```
marekmatej: `CONFIG_IEEE802154_SLEEP_ENABLE` seems to be redundant here

zephyr/port/include/zephyr_compat.h
marekmatej: maybe `CONFIG_BT_CTRL_SLEEP_ENABLE` should be aliased here too?

components/esp_phy/src/btbb_init.c
raffarost: Even though this is a new and not yet enabled symbol (by us), it's probably a good idea to include it and keep the line aligned with HAL/IDF.

zephyr/port/include/zephyr_compat.h
raffarost: As I could check, this symbol is for ESP32-H4 and ESP32-S31, so I'd add it later when it's needed/used.

sylvioalves (approved):

marekmatej (approved):

PR #651, opened 2026-09-16
clk: esp32c61: remove unused sys/cdefs.h and sys/time.h includes

kartben:
Under minimal libc `sys/time.h` falls through to the toolchain's newlib header, which needs `_BEGIN_STD_C`; the port's `sys/cdefs.h` shim does not define it, so the build stops with `expected ';' before 'typedef'`. Neither include is used by the file.

This is the same change commit ad3244bd8138 made for the seven SoCs that existed then; `esp32c61/clk.c` landed shortly before it and was missed. Fixes 15 `.minimallibc` scenarios on `esp32c61_devkitc` in the Zephyr weekly CI run.

`esp32h21`, `esp32h4` and `esp32s31` carry the same two lines (plus an unused `sys/param.h`) but Zephyr does not build those SoCs yet, so I left them out. Happy to fold them in if you prefer.


sylvioalves (approved):

sylvioalves:
@kartben yea, forgot to remove, perhaps some test could fail in CI to catch this before weekly CI.

raffarost (approved):

PR #652, opened 2026-09-16
pm: sleep: Keep MPRV off until PMP registers are restored on wake

raffarost:
The CPU loses its PMP registers when its power domain is switched off in light sleep. On wake rv_core_critical_regs_restore puts mstatus back first, which turns MPRV on again, and MPRV makes the CPU check every load and store against the now empty PMP table. The next load is refused, and since mtvec is not restored yet the trap cannot be handled, so the chip resets.

Clear MPRV in the value written to mstatus. The real mstatus is written again later by esp_sleep_cpu_retention(), after the PMP registers have been put back, so protection is re-armed as soon as it is safe.


tmedicci (approved):
Nice catch, @raffarost !

Although it doesn't seem to affect ESP-IDF, please submit this change there as well.

sylvioalves (approved):

PR #655, opened 2026-09-18
wpa_supplicant: classify sae akms without wpa3 build support

### LINKED ISSUE #119446: drivers: wifi: esp32: WPA3-only AP is reported as OPEN when CONFIG_ESP32_WIFI_ENABLE_WPA3_SAE is disabled
### Describe the bug

`CONFIG_ESP32_WIFI_ENABLE_WPA3_SAE` has no default, so it is off unless enabled. With it off, a scan reports a WPA3-Personal-only AP (SAE, no PSK AKM) as `WIFI_SECURITY_TYPE_NONE`. The shell prints that as `OPEN`. With the option on, the same AP is reported as `WIFI_SECURITY_TYPE_SAE` (`WPA3-SAE-HNP`).

An unsupported network should show as unsupported or unknown. It should not look like an open network.

Source path on main:

1. The SAE AKM selectors are only recognised under `CONFIG_WPA3_SAE`:
   https://github.com/zephyrproject-rtos/hal_espressif/blob/0db06f1fc8731e9e8b5069991744172bf9013ea4/components/wpa_supplicant/src/common/wpa_common.c#L342-L349
2. So an RSN IE whose only AKM is SAE parses with `key_mgmt == 0`, and the function still returns success:
   https://github.com/zephyrproject-rtos/hal_espressif/blob/0db06f1fc8731e9e8b5069991744172bf9013ea4/components/wpa_supplicant/src/common/wpa_common.c#L527-L538
   https://github.com/zephyrproject-rtos/hal_espressif/blob/0db06f1fc8731e9e8b5069991744172bf9013ea4/components/wpa_supplicant/src/common/wpa_common.c#L590
3. The parse callback registered with the Wi-Fi library passes `key_mgmt = 0` and `ret = 0` through:
   https://github.com/zephyrproject-rtos/hal_espressif/blob/0db06f1fc8731e9e8b5069991744172bf9013ea4/components/wpa_supplicant/esp_supplicant/src/esp_wpa_main.c#L282-L298
4. The library then reports `WIFI_AUTH_OPEN`. The driver maps that to `WIFI_SECURITY_TYPE_NONE`:
   https://github.com/zephyrproject-rtos/zephyr/blob/c605ea46e15fb9bdd518670258a8a52d8850d8f4/drivers/wifi/esp32/src/esp_wifi_drv.c#L449-L497

Step 4's authmode decision is inside the binary library. It is inferred from the A/B below, not read from source.

### To Reproduce

1. Configure an AP for WPA3-Personal only (SAE, PMF required).
2. Build a Wi-Fi application for `esp32c6_devkitc/esp32c6/hpcore` with `CONFIG_WIFI_ESP32=y`. Leave `CONFIG_ESP32_WIFI_ENABLE_WPA3_SAE` unset.
3. Scan (`NET_REQUEST_WIFI_SCAN`). The AP's result has `security = WIFI_SECURITY_TYPE_NONE`.
4. Rebuild with `CONFIG_ESP32_WIFI_ENABLE_WPA3_SAE=y` and scan again. The same AP now has `security = WIFI_SECURITY_TYPE_SAE`.

The results were read from the scan result callback. The `wifi scan` shell command prints the same field, but it was not run separately.

### Expected behavior

With SAE compiled out, a network whose only AKMs are SAE variants is reported as `WIFI_SECURITY_TYPE_UNKNOWN`, or as SAE so the application can refuse it as unsupported. It is never reported as `WIFI_SECURITY_TYPE_NONE`.

Possible directions (not tested):
- recognise the SAE AKM selectors for classification regardless of `CONFIG_WPA3_SAE`, or
- return an error from the RSN parse when an RSN IE yields no recognised AKM, and check how the library treats that result.

### Impact

- Default configuration is affected, because the option is off by default.
- An application that picks security from scan results sees "open". It can then attempt an unauthenticated association with a WPA3 network, or show it to a user as open.
- The real cause (WPA3 not built in) is hidden. The failure looks like a passphrase problem or an AP problem.
- What a `WIFI_SECURITY_TYPE_NONE` connect does against such an AP was not tested.

### Logs and console output

Four scans, same board, one config symbol changed:

| AP configuration | `ESP32_WIFI_ENABLE_WPA3_SAE` | reported security |
|---|---|---|
| WPA3-Personal only | n | `OPEN` (wrong) |
| WPA3-Personal only | y | `WPA3-SAE-HNP` |
| WPA2/WPA3 transition | n | `WPA2-PSK` |
| WPA/WPA2 mixed | n | `UNKNOWN` (mapping since added on main) |

Related, and already fixed on main: on v4.4.0, requesting a SAE connect with the option off returned `-EINVAL` and left the driver in CONNECTING. Every later connect returned `-EALREADY`. #114734 fixed that.

### Environment

- OS: Windows 11 (host)
- Toolchain: Zephyr SDK 1.0.1 (`riscv64-zephyr-elf-gcc`)
- Commit SHA or Version used: Zephyr v4.4.0 (`684c9e8f32e4`), hal_espressif `b7953b8019361d09e613f7011d2ccc41b984d087`, on ESP32-C6. The source path above was re-checked on zephyr `c605ea46e15f` and hal_espressif `0db06f1fc873`. It was not re-run on hardware at those revisions, and main pins a newer esp32-wifi-lib.

### Additional context

- The code is in hal_espressif, which has issues disabled, so this is filed here.
- Not a duplicate of #105733 (SAE build/crash). #115368 and #116038 cover transition and mixed authmodes in the same switch, a different case.

sylvioalves:
The SAE AKM selectors were only recognized when CONFIG_WPA3_SAE was set. The Wi-Fi library calls this parser to classify scan results, so a WPA3-Personal-only RSN IE yielded key_mgmt = 0 and the scan record kept authmode WIFI_AUTH_OPEN, reporting a WPA3 network as open to applications.

Recognizing an AKM selector is classification, not capability, so move the three SAE selectors out of the CONFIG_WPA3_SAE guard. The key management values are plain bit macros, so no SAE code is pulled in, and connection attempts stay gated by the driver.

ESP-IDF fixes this with a dedicated scan-only parser registered in a wpa_funcs entry that the current Wi-Fi libraries do not call, so this change is dropped once a HAL sync picks up both.

Fixes https://github.com/zephyrproject-rtos/zephyr/issues/119446


raffarost:
This might explain the issue with C5 runner auth in Gitlab/CI. Worth testing and possibly adding a wpa3 auth testcase.

raffarost (approved):

sylvioalves:
@marekmatej @tmedicci PTAL

tmedicci (approved):

marekmatej (approved):

PR #656, opened 2026-09-19
esp_hw_support: use the driver's own lock in regi2c, not irq_lock()

kartben:
The four `regi2c_ctrl_*` accessors guard their register sequences with Zephyr's
`irq_lock()`, but they run from clock/PLL bring-up in `boot_start()` — before the
kernel has a current thread, so that is not a guarantee the caller is in a
position to ask for. Use the file's own `regi2c_lock` with
`esp_os_enter_critical()` instead: it is already declared here and used by
`regi2c_enter_critical()` just below, it is CAS-based so it needs no current
thread, and for a shared analog register block a lock on the block is more
appropriate than a blanket interrupt disable.

On the single-core builds supported today this is latent — `irq_lock()` reduces
to `arch_irq_lock()`, which needs no kernel context. It becomes an obstacle to
adding SMP support for these SoCs: with `CONFIG_SMP`, `irq_lock()` is
`z_smp_global_lock()` and dereferences the current thread, so the first analog
register write during clock setup takes a NULL pointer fault. On ESP32-S3 that
is a reset loop with a saved PC in `_DoubleExceptionVector` and no output at
all, since it happens long before there is a console.


sylvioalves:
@kartben #644 already covers this. Preparing the S3 and P4 SMP PR on Zephyr side.

kartben:
> @kartben #644 already covers this. Preparing the S3 and P4 SMP PR on Zephyr side. 

Gotcha, will close then. And yeah I also have SMP running on S3 :)

